PR #279, opened 2014-12-31
Alternate landing page

dezzie:
Changed the name to index2 to prevent any overwriting. We can use my fork and then delete it after testing instead of messing with the main repo.

Sound good? @ascott1 @stephanieosan or whomever else accepts these?
...note I'm submitting this from my personal lappy, and I just upgraded to 10.10 and apparently it screws up your dev environment. If either of you got around that upgrade fine, I need to know how you got around it!! 

fatbutt


ascott1:
selfie-1

PR #285, opened 2015-01-06
Rate checker links have been removed

mistergone:
Links and content removed in a temporary fashion.

Also, this removes the debounce requirement because that change was sitting in my code. It should have been included with https://github.com/cfpb/owning-a-home/pull/265 but it wasn't... debounce isn't being used anywhere that I can see, but feel free to double-check before merging!


ascott1:
:+1:

PR #286, opened 2015-01-06
Removed index2.html

mistergone:
:no_entry_sign:


ascott1:
I'm simultaneously smart and stupid. We don't actually have to do this as the grunt copy file prevents these sorts of things from being included in a release. Sorry!

PR #292, opened 2015-01-08
Fixed email submit form + Force cache miss for css and js files

akiany:


src/_layouts/base.html
```diff
@@ -31,7 +31,7 @@
 <link rel="apple-touch-icon" size="114x114" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-114x114-precomposed.png">
 <link rel="apple-touch-icon" size="144x144" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-144x144-precomposed.png">
 
-<link rel="stylesheet" href="{{ url_for('static', filename='css/main.css') }}">
+<link rel="stylesheet" href="{{ url_for('static', filename='css/main.css') }}?v1">
```
ascott1: Will we just have to manually change these version numbers each time we push to prod?

src/static/js/modules/nemo.js
```diff
@@ -130,7 +132,7 @@ $(document).ready(function (){
                     else {
                         thanks = "Thanks, weâ€™ll be in touch.";
```
ascott1: I know this was already existing but does the "we'll" format ok? Looks like it's encoded oddly.

src/_layouts/base.html
akiany: Ideally, they would get dynamically inserted from the version number of the release as opposed to "v1". I will work on something like that, but for this release this will serve its purpose.

src/static/js/modules/nemo.js
akiany: Good catch. I had figured there was a reason for it, and decided not to touch it.

src/_layouts/base.html
ascott1: I'm cool with that for now. Let's just make sure it gets added to a backlog so that this doesn't get overlooked in the next release.

PR #293, opened 2015-01-09
Fixed the timestamp issue on Safari 8

akiany:


ascott1:
I pulled this down and tested it locally. I can confirm that it doesn't cause any problems in other browsers. Well done @akiany!

PR #300, opened 2015-01-13
Sending pass/fail to saucelabs fails if using a proxy for connection

Ooblioob:
Fixing a bug where the final step to send pass/fail to saucelabs will not work if using a proxy. It will now check if a proxy is present (via the HTTP_PROXY environment variable) and use that if it's present.  Otherwise it'll use the regular direct connection

@OrlandoSoto Can you please review this?


OrlandoSoto:
awesome! thanks @Ooblioob !

PR #301, opened 2015-01-14
reworked the email signup test and tagged smoke test scenarios

OrlandoSoto:
- reworked the test to verify the signup button disappears after clicking it
- tagged smoke test scenarios


mistergone:
:+1:

PR #303, opened 2015-01-15
Fix for arrow keys in credit score slider

mistergone:
The page will now update when the user hits the arrow keys in the credit score slider.


ascott1:
:metal: good catch

PR #304, opened 2015-01-21
Loan types final content

stephanieosan:
Update to the first paragraph of rate checker.


stephanieosan:
@wernerc @ascott1 Please review and merge!

PR #312, opened 2015-01-26
Fixes #311 - updates virtualenv variable

### LINKED ISSUE #311: In Readme, consider naming virtualenv `OAH`
Based on a discussion with @ascott1, consider renaming the virtualenv in the project Readme to `OAH` instead of `sheer` to better align the public instructions of the project with the internal instructions:

So:

```
$ mkvirtualenv sheer
```

Would become:

```
$ mkvirtualenv OAH
```

anselmbradford:
Fixes #311 - updates virtualenv variable from `sheer` to `OAH` to be
consistent with internal and external docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ Check out the sheer GitHub project (you may want to first navigate to your proje
 $ git clone https://github.com/cfpb/sheer.git
 ```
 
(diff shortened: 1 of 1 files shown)


ascott1:
Good call!

PR #324, opened 2015-02-04
Form explainer bug fix. Splits data files for easier editing.

himedlooff:
This should be an easy one to review.


himedlooff:
Tests are passing!

ascott1:
:+1:

himedlooff:
Thanks! Although I don't think this really fixed my bug = /   ... not all the way at least.

PR #325, opened 2015-02-05
Adds initial preparation worksheets app

anselmbradford:
- Updates cf-core to 0.90 from 0.5.0.
- Updates cf-forms to 1.2.0 from 1.1.0.
- Adds cf-layout 0.2.2.
- Adds `brand-palette` and `cf-theme-overrides` as needed by the latest
cf-core.
- Creates initial home buyer’s preparation worksheets HTML/CSS/JS.
Features of this commit:
- Load first user input worksheet (lifestyle goals).
- Load default goals.
- Methods for saving and loading state of inputs as JSON.
- User ability to Add/Edit/Delete goals.
- Ability to navigate to the summary worksheet, which displays output
of user inputted goals.


anselmbradford:
Looks like:

worksheet-interactive

**Note:** Most significant to OAH overall is that `cf-core` is updated in this PR! Be on the lookout for any anomalies in the rest of the site.

himedlooff:
I started looking through this and there's good stuff in here! Will make some comments.

himedlooff:
Was there a reason you couldn't use the `__disabled` and `__secondary` button styles instead of coming up with new ones? This might be more of a design question, but from the front-end and design perspective we want to keep the number of different button styles to a minimum. If we need them then that is cool. In that case I would recommend naming them something like `.btn__tertiary` so that they could be easily moved to Capital Framework.

src/static/js/templates/prepare-worksheets/input-graded.hbs
```diff
@@ -0,0 +1,15 @@
+<section class="input-graded block__border-bottom u-mb20">
+  <div class="u-mt20 u-mb20">
+    <section class="input-with-btns">
+      <div class="input-with-btns_btns">
+        <button class="btn btn__grouped-first {{#if high_grade}}active{{/if}}">High</button>
+        <button class="btn btn__grouped {{#if medium_grade}}active{{/if}}">Med</button>
+        <button class="btn btn__grouped-last {{#if low_grade}}active{{/if}}">Low</button>
+        <span class="btn-delete cf-form_input-icon cf-icon cf-icon-delete-round" role="button" title="Delete"></span>
```
himedlooff: Why not use a `<button>` here too instead of `<span>`?

src/prepare-worksheets/index.html
```diff
@@ -0,0 +1,55 @@
...
+              "grade": 2
+            }]
+        </script> 
+        
+        <section class="worksheets block block__padded-top block__border-top">
+          <!-- Content filled in dynamically from JS in prepare-worksheet.js -->
+        </section>
+
+        <nav class="block block__padded-top block__border-top">
+          <button class="btn btn-next">
```
himedlooff: I have a feeling that `.btn-next` is a little too generic. This seems to be very specific to your worksheet page, what do you think about `worksheets-next` instead?

src/static/js/templates/prepare-worksheets/input-graded.hbs
```diff
@@ -0,0 +1,15 @@
...
+      </div>
+      <div class="input-with-btns_input">
+        <input type="text" value="{{input_value}}" placeholder="Fill in your own goal">
```
himedlooff: For accessibility reasons you should include a `<label>` for the input and the buttons. I would use the text from the headers: "Lifestyle goal" and "Priority level". Then hide them with `.u-visually-hidden`. Screen readers will pick them up.

src/static/js/templates/prepare-worksheets/worksheet-goals.hbs
```diff
@@ -0,0 +1,18 @@
+<!-- Goals -->
+<h2><span class="cf-icon cf-icon-approved-round"></span> Clarify your goals</h2>
+
+<p class="lead">Most people are interested in buying a home for a mix of lifestyle and financial reasons. Make a list of all the reasons you’re interested in buying a home. <em>We’ve listed some common ones to get you started – edit the items or delete ones that are not relevant to you, and add your own item.</em></p>
+
+<!-- Lifestyle Goals -->
+<section class="worksheet worksheet-goal">
+  <header class="subheading">
+    <h4 class="u-right">Priority level</h4>
```
himedlooff: I would recommend using `aria-hidden` to keep these from getting read by screen readers. I don't think it would make sense to hear these two headings read out. The `<label>`'s from my previous comment replace the need for these anyway.

src/static/css/module/prepare-worksheets.less
```diff
@@ -0,0 +1,112 @@
...
+  width: 100%;
+}
+
+.input-with-btns {
+  display: block;
+  clear: both;
+  margin-bottom: 10px;
+}
+
+// Preparation worksheets goals, risks, red flag icon colors.
```
himedlooff: **Suggestion:** Better selectors for these might be `.worksheets-icon__goals, .worksheets-icon__risk, .worksheets-icon__flag`. They are more declarative and decouples the color from the icon. For example if you wanted to use the non-round versions or a different icon altogether, you can do so without having to change the CSS.

src/static/css/module/prepare-worksheets.less
```diff
@@ -0,0 +1,112 @@
...
+
+.prepare-worksheet .cf-icon-error-round {
+  color: @gold;
+}
+
+// Inputs with buttons for goals, risks, and red flags.
+// The color settings override CF colors.
+
+// Right-hand column of the worksheets (priority, likely-ness, etc).
+.worksheet .u-right {
```
himedlooff: **Suggestion:** It would be nice if this was more declarative. `.worksheets-header-col` or something. This also has the benefit of allowing you to add `.u-right` in the CSS instead of the HTML:

```less
.worksheets-header-col {
  .u-right();
  width: @right-column-width;
}
```

```html
<h4 class="worksheets-header-col">Priority level</h4>
```

src/static/js/templates/prepare-worksheets/worksheet-summary.hbs
```diff
@@ -0,0 +1,22 @@
...
+<h2><span class="cf-icon cf-icon-information-round"></span> Summary</h2>
+
+<!-- Lifestyle Goals -->
+<section class="worksheet worksheet-goal">
+  <header class="subheading">
+    <h4 class="u-right">Alternative</h4>
+    <h4>Lifestyle goal</h4>
+  </header>
+
+  <section class="text-group">
```
himedlooff: Is this class being used anywhere else?

src/static/js/templates/prepare-worksheets/worksheet-summary.hbs
```diff
@@ -0,0 +1,22 @@
...
+    {{#each inputs}}
+    <section class="input-graded block__border-bottom u-mb20">
+	  <div class="u-mt20 u-mb20">
+	    <section class="input-summary">
```
himedlooff: Is this class being used anywhere else?

anselmbradford:
> Was there a reason you couldn't use the __disabled and __secondary button styles instead of coming up with new ones? 

Mostly unfamiliarity with the codebase ;) Can you clarify actually which buttons you mean are new? Do you mean the `btn-delete` and `btn-add-input-graded` classes or do you mean the color overrides I added? 
Also, as far as `__disabled` ... I don't think that would be applicable in this case, because none of the buttons are truly disabled (meaning the user can't click them), they're just inactive and active.

> Same with .btn-delete. This could be a useful CF pattern. Maybe something like .icon-btn.

It might actually be useful to have a btn class in CF that is devoid of styles so that it is semantically a button, but doesn't have the default border and background. An icon or any other content could then be included inside the button. I'm not sure if there is a preference for explicitly creating a class for a button that contains an icon versus creating a button class that is devoid of default styling, which could contain an icon, if needed. Thoughts?

src/static/js/templates/prepare-worksheets/input-graded.hbs
anselmbradford: I agree `<button>` would be preferable here. I did this because the design was in flux and using a button would mean I would have to override all the default styles (see https://github.com/cfpb/owning-a-home/pull/325#issuecomment-73053921). I wanted to keep the stylesheet at a minimum of customized overrides while developing, but yes, this should eventually be a button.

As a side note, I actually wonder if `<button>` and `<span role='button'>` are semantically equivalent? Do you know?

himedlooff:
> Can you clarify actually which buttons you mean are new?

The buttons that seem new are the light gray and dark gray ones. Basically, the buttons that you needed to override colors for. Our two main styles are:

image

Good point about the `__disabled` button style, it does come along with that disabled cursor. Also, we might not want to use our disabled style in place of a non-disabled button anyway.

himedlooff:
> It might actually be useful to have a btn class in CF that is devoid of styles

We have `.btn__link`, which has no bg color and instead colors the text. It does have a bottom border though. Maybe there could be a version of this with no border?

src/prepare-worksheets/index.html
anselmbradford: > I have a feeling that .btn-next is a little too generic. This seems to be very specific to your worksheet page, what do you think about worksheet-next instead?

So the difference between `btn-next` and `worksheet-next` that I see is `btn-next` describes the element+action, `worksheet-next` describes the application+action. Is there a precedent for which is preferred in the project? Also, is this the preferred workflow in the project for identifying a button for adding events (e.g. adding a custom class)?

src/static/js/templates/prepare-worksheets/input-graded.hbs
himedlooff: Sounds good! Not sure if `<button>` and `<span role='button'>` are semantically equivalent.

src/prepare-worksheets/index.html
himedlooff: I recommended this because it seems you're only using this class as a js-hook for your application.  If it's application-specific stuff I think it's best to name it that way. Others who come along looking at the code can more easily identify what pieces are more application-specific.

src/static/js/templates/prepare-worksheets/input-graded.hbs
anselmbradford: Sounds good :+1: Alternatively, I had "Lifestyle goal" and "Priority level" inside `<label>`'s originally, that would work also, right? 

Question about the priority label—Since the `for` attribute of a label means the label can be associated with a single form control, should a duplicate label be created for each button, or can screen readers handle something like, e.g.:

```
<label><span class='.u-visually-hidden'>Priority level</span>
    <button class="btn btn__grouped-first">High</button>
    <button class="btn btn__grouped">Med</button>
    <button class="btn btn__grouped-last">Low</button>
</label>
```

src/static/js/templates/prepare-worksheets/worksheet-goals.hbs
anselmbradford: :+1:

src/static/js/templates/prepare-worksheets/input-graded.hbs
himedlooff: Wrapping the buttons in a label is fine. I wonder if we need to clarify in the label that you should choose one out of the 3? Or maybe just "Priority level: choose one"?

himedlooff:
@anselmbradford, I made a bunch of comments/suggestions. I know this is a WIP so let me know when you want me to merge so you can forge ahead.

anselmbradford:
> The buttons that seem new are the light gray and dark gray ones.

For background, see [this issue](https://github.cfpb.gov/OAH/OAH-notes/issues/770). Maybe there should be button styles specific to the grouped buttons, since those can have styles that depend on the state of other buttons in the group.

> We have .btn__link, which has no bg color and instead colors the text. It does have a bottom border though. Maybe there could be a version of this with no border?

Would `.btn__nostyle` work?

>  I made a bunch of comments/suggestions. I know this is a WIP so let me know when you want me to merge so you can forge ahead.

No prob, I should have clarified in the PR that this was mostly for my team to be able to play with the current state of the project on build. I'll check in my standup if they want it merged before addressing the issues. Your comments help me learn the codebase practices, so thanks for putting in the review time.

himedlooff:
> Maybe there should be button styles specific to the grouped buttons

Those dark gray buttons look very similar to `__secondary` so I'd use those. Then you'd only need the custom navy blue "active" button. Maybe `.btn__grouped-active` would work for a generic name.

himedlooff:
> Would .btn__nostyle work?

Not sure. For your needs the icon should be kind of big right? If that's the case maybe it is icon-specific like `.icon-btn`?

himedlooff:
> Your comments help me learn the codebase practices, so thanks for putting in the review time.

No problem. Glad you're finding it helpful, that was the goal. It's also helpful for me.

himedlooff:
> Maybe there should be button styles specific to the grouped buttons

Oops, didn't scroll down far enough. Those aren't `__secondary`.

Scotchester:
This review and discussion warms the cockles of my heart :smile:

himedlooff:
Talked to @anselmbradford and I'm good to merge. @ascott1, @contolini, did you want to take a look at all?

anselmbradford:
Yeah, this is a work-in-progress for my team to play with the prototype on build, as it is. Please feel free to note any code quality issues, which I'll pick up for another PR or pass off to the OAH team after small projects end.

anselmbradford:
This should end up at http://build.consumerfinance.gov/prepare-worksheets/, amirite?

PR #333, opened 2015-02-07
Tests for a couple more modules

contolini:
Some unit tests for the dropdown-utils and humanize modules. Dropdown-utils needs to be extracted and published as a standalone node module at some point.


ascott1:
Coveralls you make no sense.

PR #340, opened 2015-02-14
the links recently added now open up in new tabs

stephanieosan:
When I added the links in the intro paragraph of the rate checker several weeks ago, I forgot to make those links open in a new tab. This fixes that.

Also, stuck in the airport because my plane had a crack in the windshield. ARGH.


stephanieosan:
selfie-0

cfarm:
:+1:

PR #342, opened 2015-02-17
Create empty config file if it doesn't exist

contolini:


cfarm:
Looks like duplicate of https://github.com/cfpb/owning-a-home/pull/344

PR #344, opened 2015-02-18
Taking steps toward ci and shorter dev feedback loop

imuchnik:
grunt build now runs tests and coverage
frontendbuild.sh is now standalone and idempotent.


cfarm:
:+1:

PR #345, opened 2015-02-18
Consolidate 'prepare worksheet' editors

virginiacc:
- Combines worksheet templates & controllers
- Adds new page level templates


cfarm:
Let's get this merged so we can start workin. Would welcome @anselmbradford's feedback too

src/static/js/modules/prepare-worksheets/inputs/button-grading-group.js
```diff
@@ -11,42 +11,39 @@ function ButtonGradingGroup(options) {
   // Note bind()'s lack of IE8 support.
   var _self = this;
 
-  var _grades = require('./input-graded-grades');
-
-  // The grade of this input.
-  var grade = _grades.UNSET;
-  var lastNode;
-
+  this.row = options.row;
```
anselmbradford: Is `row` used outside of this module? e.g. is something like this code somewhere else in the codebase:

```javascript
var ButtonGradingGroup = require( 'button-grading-group' );
...
var group = ButtonGradingGroup.create( options );
console.log( group.row ); // <-- Is this happening somewhere else where there's a need to get access to the row externally to the module?
```

anselmbradford: Additional to the above, I'd suggest creating a locally scoped var and then doing all the exposing of the methods/properties at the end. So like this:

```javascript
var row = options.row;

...

  // Expose public methods
  this.getGrade = getGrade;
  this.setGrade = setGrade;
  this.row = row;
```

You may have to create a method instead of exposing the property directly, to avoid problems:

```javascript
var _row = options.row;

...
  // @return [Object] A row of data.
  function getRow() {
    return _row;
  }

  // Expose public methods
  this.getGrade = getGrade;
  this.setGrade = setGrade;
  this.getRow = getRow;
```

**Note:** I use the underscore prefix to designate what is not being exposed externally.

src/static/js/templates/prepare-worksheets/input-graded.hbs
```diff
@@ -3,16 +3,16 @@
     <section class="input-with-btns">
       <div class="input-with-btns_btns">
         <label><span class="u-visually-hidden">Priority level: choose one</span>
-          <button class="btn btn__grouped-first {{#if highGrade}}active{{/if}}">{{highGradeText}}</button>
-          <button class="btn btn__grouped {{#if mediumGrade}}active{{/if}}">{{mediumGradeText}}</button>
-          <button class="btn btn__grouped-last {{#if lowGrade}}active{{/if}}">{{lowGradeText}}</button>
-          {{#if deletable}}
+          <button class="btn btn__grouped-first {{#if is_0}}active{{/if}}">{{grades.[0]}}</button>
```
anselmbradford: Is `is_0` present in the JS anywhere or would that be passed in from an external API? If it's in the JS, run `grunt jshint`. I think it will complain about the underscore variable name.

src/static/js/modules/prepare-worksheets/inputs/button-grading-group.js
virginiacc: Good point! There's no need for that to be exposed externally, so I'll scope it locally.

src/static/js/templates/prepare-worksheets/input-graded.hbs
virginiacc: It's created as a property of the data object passed into the template, but there's really no need for the underscore. I'll remove it.

PR #352, opened 2015-02-20
Worksheet expandables

cfarm:
## Additions
* Adds cf-expandables component to JavaScript vendor modules (JS + Less)
* Adds cf-expandables component as a dependency in prepare-worksheets.js so we can use expandables
* Triggers expandables js for worksheets page load so we can use it in our JS templates
* Adds basic expandables link pattern to summary page for alternatives

## Changes from Design
* This summary page is still under development by @virginiacc to match the design comps so the only styling here is for the expandables links + paragraph content themselves. 

## Review
* @virginiacc 
* @huetingj: let me know if the open state is ok, as I didn't see one in the .ai file, ditto for the "Hide your alternative" copy

### Alternatives expandables: top is the open state, bottom is the closed state
screen shot 2015-02-20 at 4 27 50 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ If you already have these modules installed, [skip ahead to Sheer](#sheer-elasti
 	$ brew install elasticsearch
 	```
 
-2. Clone the sheer GitHub project:
+2. Clone the sheer GitHub project to wherever you keep your projects (not inside Owning a Home):
 	```
 	$ git clone https://github.com/cfpb/sheer.git
 	```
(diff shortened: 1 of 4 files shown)


cfarm:
Oh, I forgot to only output this for Goals. Update on the way....

cfarm:
OK, fixed and ready for review. Just a note that if/when Risks and Red Flags notes content gets added back to the config, they should be added as 'explanation' value instead of altText which is specifically for alternatives and which output on the summary.

virginiacc:
This looks great! I'll merge it now & we can keep the alternative design questions in mind for possible later visual qa.

huetingj:
The expandable looks great--- No changes, and thanks for formatting it.

PR #357, opened 2015-02-25
Worksheets: make default risks & flags uneditable

virginiacc:
## Additions
- add uneditable flag to default risks & flags
- show text instead of input for uneditable worksheet items

## Screenshot
screen shot 2015-02-25 at 5 45 43 pm


cfarm:
Nice job

src/static/js/modules/prepare-worksheets/worksheet-config.js
```diff
@@ -223,7 +228,8 @@ this.worksheetDefaults = {
                 "alternative":"Think of all the little things that you are used to calling your landlord to deal with: a cracked window, a broken dishwasher, or a clogged toilet. As a homeowner, you will either have to fix these yourself or call and pay for a professional.",
                 "altText": "",
                 "explanation":"",
-                "deletable": false
+                "deletable": false,
+                "uneditable": true
```
cfarm: this is a quibble but to me it's a lot of negatives and potentially more confusing to have 'uneditable' rather than an 'editable' key. would be more readable to see if it's editable then check if it's not uneditable, for example here https://github.com/cfpb/owning-a-home/pull/357/files#diff-8685b10fc3ea02fd0a2b3942e2505d01R57

virginiacc: Good point. I was taking the easy route here -- if false is the most common case, you only need to specify the attribute when true. But 'editable' is probably a better/more consistent approach. I'll change it...

virginiacc: Although now that I think about it, the inability to edit/delete are both results of the row being required. I think I'll consolidate these as one attribute.

virginiacc:
Okay, the `deletable` and `uneditable` flags are now consolidated as `required`.

src/static/js/templates/prepare-worksheets/input-graded.hbs
```diff
@@ -10,9 +15,9 @@
           <button class="btn btn__grouped-first {{#if is_0}}active{{/if}}">{{grades.[0]}}</button>
           <button class="btn btn__grouped {{#if is_1}}active{{/if}}">{{grades.[1]}}</button>
           <button class="btn btn__grouped-last {{#if is_2}}active{{/if}}">{{grades.[2]}}</button>
-          {{#if input.deletable}}
+          {{#unless input.required}}
```
cfarm: nifty.

cfarm:
Yo this makes a lot of sense :+1:

PR #355, opened 2015-02-24
First journey page & journey step template

virginiacc:
## Additions
- First journey/`know the process` page (`prepare` step)
- `.process-step` template
- `know-the-process.less`

## Removals
- duplicated `cf-typography.less` styles from `typography.less`

## Notes
- The css still needs some tweaks, but I'd like to get this out so we can start discussion
- Since it seems like each of the step pages will have similar content, I've created a shared `_process-step` template. The template can be populated with step-specific data declared in the `index.html` page for each individual step. 
- The journey naming/path conventions are still under discussion, so for now I'm putting the steps in a `process` directory with the first step in the main `index.html`.

## Screenshots
owning a home know the process gt consumer financial protection bureau

PR #358, opened 2015-02-26
Worksheet bug fixes

contolini:
This'll:
- Tell Travis CI to run unit tests against ES6-compliant io.js instead of Node.js to avoid [syntax errors](https://github.com/cfpb/owning-a-home/pull/355#issuecomment-75980176).
- Set up a test suite for the prepare-worksheets section of the site (but I didn't write any actual tests yet). I mocked the page's DOM in a kinda dumb way but because we're reliant on Sheer I couldn't think of a better method.
- Fix the expandables bug on the prepare-worksheets page. I think? I'm not totally sure what the desired functionality is so you'll want to double-check that I didn't break anything on that page.

### DIFF
--- a/test/js/prepare-worksheets_test.js
+++ b/test/js/prepare-worksheets_test.js
@@ -1,0 +1,29 @@
+var expect = require('chai').expect;
+var jsdom = require('mocha-jsdom');
+
+describe('Prepare worksheets', function() {
+
+  var $,
+      worksheets;
+
+  // Load in the HTML of the prepare-worksheets page. It'd be ideal to load the
+  // live content but Sheer makes that difficult.
+  jsdom({
+    file: 'test/js/fixtures/prepare-worksheets.html',
+    console: false
+  });
+
+  before(function () {
+    $ = require('jquery');
+    worksheets = require('../../src/static/js/modules/prepare-worksheets/prepare-worksheets.js');
+  });
+
+  beforeEach(function() {
+
+  });
+
+  it('should complain if you don\'t specify a target', function() {
+
+  });
+
+});
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '0.10'
+  - iojs
 
 install:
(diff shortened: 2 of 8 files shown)


test/js/prepare-worksheets_test.js
```diff
@@ -0,0 +1,29 @@
+var expect = require('chai').expect;
+var jsdom = require('mocha-jsdom');
+
+describe('Prepare worksheets', function() {
+
+  var $,
+      worksheets;
+
+  // Load in the HTML of the prepare-worksheets page. It'd be ideal to load the
+  // live content but Sheer makes that difficult.
```
cfarm: What's up with Sheer? Can someone make it easier for us to do that?

cfarm:
Mobile secondary nav is working now, thank you! 

Since expandables are gone for now the summary page alternatives expandable doesn't work. After this gets merge I'll fix this by removing the expandable and reverting to our MVP of just displaying the alternative text w/o hiding it inside an expandable.

test/js/prepare-worksheets_test.js
contolini: Normally we'd use something like [grunt-contrib-jasmine](https://github.com/gruntjs/grunt-contrib-jasmine) to automagically run our tests through PhantomJS but OAH can't do that because our templates are processed and served by Sheer. We have to first install and start Sheer to get the HTML of any OAH page. Locally, this isn't a problem but with external services like Travis, we'd have to write a script to get it up and running on their virtual machines. This would be difficult and it kind of defeats the purpose of unit tests which are supposed to test small, individual units of behavior.

The proper solution IMHO is to refactor the code to decouple it from the DOM. That way we can test methods' logic without requiring much HTML. The [dropdown tests](https://github.com/cfpb/owning-a-home/blob/master/test/js/dropdown-utils_test.js#L16) are a good example of this.

For now we're just pasting in a bunch of HTML and saying "Hey Mocha, pretend this huge blob of HTML is a webpage and run the prepare-worksheet tests against it".

PR #361, opened 2015-02-26
Scroll to top on worksheet page change

virginiacc:
- The animation time value may need to be tweaked


cfarm:
:+1: 
I think it's good not to make the scroll too fast but dunno if there's a standard we should be using.
This is probably a separate feature, but do we want to scroll down to flags/risks for the error messaging? eg when you click "#page1risks"

virginiacc:
It might be too fast now since I just made up the duration, but I don't know of a standard either.

I agree that it would be helpful to add scrolling after error message links. I'll try to get to that today -- if not, we can add it to the backlog.

PR #363, opened 2015-02-26
Fix hash conflict caused by summary page error links

virginiacc:
- prevent hash change on error link click -- causes conflict with mega expand


cfarm:
:+1: all better now.

PR #366, opened 2015-02-27
working sourcemaps

ascott1:
Running `grunt less:map` will now produce a working sourcemap so that you can dig into the CSS more easily from dev tools. It's not in the watch task as it is super slow (around 7 seconds - so many files!). 

The gruntfile could use some love in general. I'm happy to help with that.

Also, HI OAH!!!
selfie-0

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,7 +63,6 @@ module.exports = function(grunt) {
      * LESS: https://github.com/gruntjs/grunt-contrib-less
      *
      * Compile LESS files to CSS.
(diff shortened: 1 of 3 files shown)


cfarm:
grunthelp invite extended
I'm also adding that to the list of refactor needs.

cfarm:
BEAUTIFUL.

If you can get our Gruntfile tasks to execute more quickly, we should add this to the watch task. Watch is already slow anyway, and this is infinitely more useful while we're making changes and running watch.

PR #375, opened 2015-03-07
html/css graph (Journey Summary)

billimarie:
Hey everyone,

Here's a template of the "Journal Summary" page.

There's a placeholder for the graph you guys made in backbone.js. I left the "Alternatives" tab in but it can easily be hidden (via visibility hidden non etc.)

billimarie


billimarie:
your-summary-html-css-mockup

cfarm:
This is incorporated into #377

PR #377, opened 2015-03-07
Worksheet summary graph js

cfarm:
## Added
* Loops for outputting graph lines for goals, risks, and red flags

## Todo
* Don't output graph container if there are no results


cfarm:
@amymok ready for review

cfarm:
note on this: you can't see the code locally anymore because we removed the prepare worksheets JS from the Gruntfile so it's not compiled (because of our `grunt deploy` task errors)

cfarm:
to build this locally you can uncomment these 2 lines in Gruntfile.js

https://github.com/cfpb/owning-a-home/blob/master/Gruntfile.js#L119

https://github.com/cfpb/owning-a-home/blob/master/Gruntfile.js#L152

should be tested with grunt deploy task too to make sure that doesn't break if you decide to merge.

cc @virginiacc

amymok:
I am not seeing the summary graph coming out.
screen shot 2015-08-28 at 6 23 00 pm

PR #380, opened 2015-03-10
Update link to Tell Your Story in subnav

Scotchester:
Nothing needs to be rebuilt for this to work, right?


cfarm:
What do you mean? There is a deploy step to get this live.

Scotchester:
I don't need to run Grunt for these link changes, right? They are served directly from the files I changed?

cfarm:
Hmm, I thought you had to but I don't actually see which task would build these files in the Gruntfile. @virginiacc?

cfarm:
But no, you don't need to build anything to get these onto the dev site, to answer your question.

contolini:
Yeah, @Ooblioob can push to prod.

contolini:
Looks like browser tests in Jenkins are failing (some links where changed without updating the tests). Jenkins is configured to not deploy anything if tests are failing. So if this fix needs to be deployed ASAP, let's fix dem tests. @imuchnik and I can help.

Scotchester:
I did up date the test! Well, **a** test, at least. Check out the diff.

Is there some other place I should update tests?

cfarm:
@contolini bump

imuchnik:
your build is broken, we should talk.

PR #382, opened 2015-03-11
Loan comparison

virginiacc:
### Additions
- Displays interest rate update button when user changes loan data
- Makes api call to get interest rates on loan addition and when update button is clicked

### Changes
- Moves state & credit score selection out of individual loans to single 'onboarding' section
- Some refactoring of loan comparison page to pull in CF patterns

### Screenshots
screen shot 2015-03-11 at 2 30 05 am

### Review
- @cfarm 
- @stephanieosan


src/static/css/module/loan-comparison.less
```diff
@@ -9,84 +10,91 @@
...
+     }
+
+     .intro-img {
+       display: none; // hide the hero image on mobile
+     }
+   }
+});
+
+
+#onboarding {
```
cfarm: Can this be a class instead of an ID?

cfarm:
Slowly reviewing the JavaScript changes for this PR. I checked out the branch locally and found a bug in Chrome (untested in other browsers):

On any loan:
* Select "Adjustable" from Rate structure
* Select "15 years" from Loan term
* Click the "Update rates" button
The select that loads has no rates. If this is correct then we should have some type of error messaging for this case.

src/static/css/module/loan-comparison.less
```diff
@@ -157,73 +173,89 @@
...
-  }
-
+  text-align: center;
+  
   h3 {
     .reset;
   }
 }
 
 /* labels and inputs are full width on small screens */
```
cfarm: Can these be full width on less than 400px width screens instead of 600? They look weird between 400-600:

screen shot 2015-03-12 at 11 43 40 am

cfarm:
Also this is a design question for @stephanieosan: Why do we have the user click the button to update instead of making the call each time a button changes? Performance reasons, in case the user is making many changes at once?

src/static/css/module/loan-comparison.less
virginiacc: Yes -- I'll change that.

virginiacc:
@cfarm Good catch on the missing 15-year ARM rates. The api is not returning any rate data for that scenario. I'll log an issue about it.

mthibos:
@virginiacc 

We don't have data fro 15 year ARMs.

Which is perhaps a good time to mention that there are all kinds of logic dependencies for non-existent loan combinations that will have to be coded in eventually.  They are all the same as in rate checker, though, so hopefully we can just repurpose those modules.

mthibos:
@cfarm   yes, because any change to an input changes the *the entire page* -- it would be way too much going on such that we would expect very poor performance, and also not a great user experience.

virginiacc:
@mthibos, thanks! I see that rate checker is managing the allowable select options to prevent those non-existent scenarios. I'll look at adding that functionality here.

virginiacc:
@cfarm, I just added a few changes, including a switch to using a class instead of id for the onboarding section css. I'm working on pulling in the logic from rate checker, but I'll put that in a separate PR later so it doesn't hold this one up.

cfarm:
Looks good!

PR #388, opened 2015-03-18
Fix failing loan comparison tests

virginiacc:


cfarm:
Yay tests pass :smile_cat:

cfarm:
Now we wait 1 hour.

PR #389, opened 2015-03-19
Change Loan Comparison credit score ranges

virginiacc:
### Changes
- Use same credit score ranges in loan-comparison and check-rates so that rates returned are consistent.


cfarm:
:+1: Getting the same rates as Rate Checker now

stephanieosan:
awesome!!

PR #393, opened 2015-03-20
Limit files in less concat task to prevent css duplication

virginiacc:
### Changes
- Stops concatenation of `cf-concat` & most `cf-less` files.

### Review
- @cfarm


cfarm:
YAY! Trying this out now.

cfarm:
Success! No more duplicate CSS + IE9 doesn't die from selector overload anymore :dancer:

PR #395, opened 2015-03-30
Loan options expandables

cfarm:
## Changes
* Removes `mega_expand` js and less files 
* Replaces loan options `mega_expand` functionality with CF expandables
* Markup, JS, and CSS added to get CF expandables working on loan options page
* Browser tests updated for loan options page to look for new expandables selectors (many thanks to @OrlandoSoto for helping fix these!)
* Secondary navigation for mobile uses cf-expandables, markup was refactored to work with the latest cf-expandables pattern
* dependencies for secondary navigation were updated in various JS modules
* changes to dependencies caused the `prepare-worksheets.js` test to fail, so I updated its jQuery dependency to fix
* CSS was refactored for the loan options page to remove duplicate code, mega_expand related styles
* Fixed loan options index page template to match the cf.gov widths and also have proper small screen whitespace, see updated screenshot

## Review
* @virginiacc 

## Screenshots
These should look the same as the current live loan options page, save some minor expandables-related design pattern updates that are inherited from Capital Framework:

screen shot 2015-03-30 at 5 29 17 pm
screen shot 2015-03-30 at 4 48 58 pm


virginiacc:
This looks good! Although when I run the browser tests, I'm seeing issues with some of the links inside the expandables not being clickable (sample error message: Element is not clickable at point (499, 373). Other element would receive the click.) 

One of the suggestions in [this](https://code.google.com/p/selenium/issues/detail?id=2766) thread is to use a more targeted click approach, and the issue seems to be fixed by replacing the `element.click()` call in the `navigation.py` `click_link` method with:

```python
    action = webdriver.ActionChains(self.driver)\
                  .move_to_element_with_offset(element, 0, 20) .click()
    action.perform()
```
I need to run the full test suite to make sure this doesn't break anything else though.

cfarm:
We could also create our own method for this so it doesn't break other tests using `click_link` - I can give that a try.

virginiacc:
@cfarm That sounds good!

cfarm:
@virginiacc can you confirm which browser tests you've seen fail? I just ran the following on this branch and they passed:

`behave -k -t=~prod_only -t=loan_options_expandable`
`behave -k -t=~prod_only features/loan_options.feature`

virginiacc:
@cfarm, I was running the full `loan_options_expandable` test suite, just to make sure that the `prod_only` tests wouldn't fail in unexpected ways. 

The problem I saw was with the `Scenario Outline: Expand 'Loan Types' section then click Non-Navigational links inside the expanded section` tests. I expected those to fail at loading the designated url, but they were failing at the `And I click on the "<link_name>" link` step.

cfarm:
I just got this error on the following link only:

`And I click on the "specific rules" link`

`Element is not clickable at point (1034, 258). Other element would receive the click: <p>...</p>`

Will investigate your fix for this

cfarm:
@virginiacc I got your fix working locally on the following tests:

`behave -k -t=loan_options_expandable`
`behave -k features/navigation.feature`
`behave -k -t=prod_only`

BUT now they fail on the next step, and report `tab not found.` @OrlandoSoto I tried increasing the wait time for this step (from 5 to 10) in case that's the problem but it didn't help. Thoughts?

OrlandoSoto:
@cfarm 'tab not found' is reported when the test clicks on a link but doesn't find the link open in a new browser tab. 
Which scenario failed? I'm gonna try to see if I can reproduce it locally

cfarm:
@OrlandoSoto I get `tab not found` when running:

`behave -k -t=loan_options` 

for all the relative links AND for the external links

OrlandoSoto:
@cfarm, I just ran this branch,
The test goes to: 
- http://localhost:7000/loan-options/conventional-loans/

Then it clicks on Other Loan Types - FHA loans
This link opens in the same window
- http://localhost:7000/loan-options/FHA-loans/ 

But the test is expecting it to open in a new tab
Did something change to cause to cause those links to open in the same tab?

cfarm:
aha, thanks! they are missing target="_blank"

OrlandoSoto:
it's great to see the browser tests catching those changes

stephanieosan:
Hey! Just checking, are the links in question these links, the ones visible when you open up the "Loan Types" expandable?

screen shot 2015-04-08 at 3 40 36 pm

If so, they shouldn't open in a new tab, since those links are navigating you deeper into this section of the site.

cfarm:
thanks @stephanieosan let me double check these test case scenarios to make sure that's what we're expecting.

cfarm:
For the new tab errors, here's what I see happening in the browser on the failing `behave -k features/loan_options.feature` test:

* It opens the URL in a new tab. The URL usually 404s, then it seems to open the loan options index page in the same new tab again.
* This continues repeatedly 

Some of the URLs never fail in this way, though they should be using the same steps and they have the same markup. One that I don't see failing is `how-do-mortgage-lenders-calculate-monthly-payments.html`

cfarm:
With or without the fix I am getting the 'tab not found' error on this test:

`behave -k features/loan_options.feature`

@OrlandoSoto @virginiacc can one of you try to replicate?

OrlandoSoto:
@cfarm ok, I'm gonna try it, loan-options-expandables branch right?

cfarm:
Yes, thanks!

OrlandoSoto:
I ran: behave -k features/loan_options.feature -t=loan_options_expandable
and I saw a couple different type of errors:
- WebDriverException: Message: u'unknown error: Element is not clickable at point (853, 426). Other element would receive the click: <section class="col-4 sans">...</section>
- Expected: a string containing u'What is a Qualified Mortgage?'
           but: was u'/askcfpb/1789/what-qualified-mortgage.html tab not found!' 
Is that also what you saw?

cfarm:
Yes: we fixed the first error though (if you pull the latest from this branch you should have it) and I'm trying to fix the second now

cfarm:
All the tab not found failures I'm getting are for properly marked up links like the following:

`<a href="/askcfpb/1957/what-is-a-prepayment-penalty.html" target="_blank">pre-payment penalties</a>`

virginiacc:
@cfarm, I'm seeing the same results -- `tab not found` errors for links that open in a new tab, with or without the clickable element fix.

OrlandoSoto:
when I run this manually:
- Given I navigate to the "Loan Options" page                                                                              (http://localhost:7000/loan-options)
- When I click Learn More to expand the "Interest rate type" section                                                                                 
- And I click on the "pre-payment penalties" link
- Then I should see the "/askcfpb/1957/what-is-a-prepayment-penalty.html" URL with page title Consumer Financial Protection Bureau open in a new tab

I get "Please provide a 404.html!"
So that would cause the tab not found error

cfarm:
Why should that cause a different error than the following that I usually see for local URLs that don't work:

Then I should see the "/askcfpb/1941/on-a-mortgage-whats-the-difference-between-my-principal-and-interest-payment-and-my-total-monthly-payment.html" URL with page title Consumer Financial Protection Bureau open in a new tab # features/steps/steps_navigation.py:89 0.132s
      Assertion Failed: 
      Expected: a string containing u'Consumer Financial Protection Bureau'
           but: was u'http://localhost:7000/askcfpb/1941/on-a-mortgage-whats-the-difference-between-my-principal-and-interest-payment-and-my-total-monthly-payment.html'

cfarm:
I also get that error on valid local URLs:

Then I should see the "/check-rates/" URL with page title Consumer Financial Protection Bureau open in a new tab # features/steps/steps_navigation.py:89 0.023sts: ] 2015-04-09 12:38:38,627 - INFO - Current url: http://localhost:7000/loan-options/
      Assertion Failed: 
      Expected: a string containing u'Consumer Financial Protection Bureau'
           but: was u'/check-rates/ tab not found!'

cfarm:
OK, sounds like this error is expected

Merging because @OrlandoSoto said "I'm sure it will pass on build" :smiley_cat:

PR #396, opened 2015-03-31
Journey nav

virginiacc:
### Changes
- Restyles phase to match new designs
- Automatically applies styles to step `what to know` and `how to take action` sections
- Pulls in WP content for each of the journey phases

### Additions
- Top and bottom step nav sections

### Screenshots

screen shot 2015-03-31 at 4 21 30 am

screen shot 2015-03-31 at 4 21 44 am

### Review
@cfarm 
@huetingj 
@stephanieosan

### DIFF
--- a/src/static/css/module/process.less
+++ b/src/static/css/module/process.less
@@ -90,37 +90,194 @@
 
 .know-the-process {
   .process-nav {
-
-  .respond-to-min(600px, {
-    .content-l;
-    text-align: center;
-
-  });
-  .respond-to-max(599px, {
-    .jump-link;
-  });
-
-  &__top {
+    counter-reset: li;
     li {
-      .respond-to-min(600px, {
-        .grid_column(1,6);
+      br {
+        display: none;
+      }
+
+      &.process-nav_header {
+        background-color: @gray-10;
+        border-color: @gray-10;
+      }
+    }
+
+    li .process-nav_number {
+      padding-left: 1.5em;
+      position: relative;
+      margin-bottom: 5px;
+      
+      &:before {
+        .icon-round;
+        content: counter(li);
+        counter-increment: li;
+        background: @green-tint;
+        color: @black;
+        font-size: .85em;
+        .webfont-demi();
+      }
+      
+      .respond-to-min(@tablet-min, {
+        padding-left: 0;
+        margin-top: 1.75em;
+        &:before {
+          margin-left:-15px;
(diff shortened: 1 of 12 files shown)


Scotchester:
Lookin' sharp! :hocho:

Is the variance in link colors due to inheriting the standard link colors? Should we disable the `:visited` color change for a case like this?

stephanieosan:
I agree, looks great! I would be on board with disabling the visited color change. Is that what y'all do in other navigation elements in Flapjack?

Also, this feedback isn't about the nav, but that phase intro paragraph ("Buying a home...") should be Avenir instead of Georgia. @huetingj What size did you have that at—18? 22?

huetingj:
Yes--- Avenir 22 pt. And agreed about disabling the color change!

Scotchester:
I believe that's our standard practice in Flapjack, yes.

Scotchester:
The only thing that gives me pause about suggesting it here is that this navigational element also serves as an indicator for your position in a multi-step process. Maybe there's value in indicating which steps you've already visited?

stephanieosan:
Hmm, that's a good point. Maybe @emiliecfpb or @sonnakim could weigh in on which is preferred?

src/static/css/module/process.less
```diff
@@ -67,58 +67,304 @@
...
+  .webfont-demi();
+  float: left;
+  margin-right: 5px;
 }
 
 /* Process-specific styles */
 
 .know-the-process {
   
+  .process-nav {
```
cfarm: does this need to be nested in `.know-the-process` since you already have a class specific to this nav?

cfarm:
This looks good, the styling for the WYSIWYG looks great too - a minor CSS update might be in order but once that is done I'd merge.

src/static/css/module/process.less
virginiacc: Good point! I've updated this.

virginiacc:
Okay, I've fixed the font on the intro paragraph:
screen shot 2015-04-01 at 1 27 01 pm

and decreased the nesting on the journey nav css.

I agree about the color variation on the nav links. Although I wonder if, as @Scotchester suggests, it might make more sense as a progress tracker if you were going through the steps in order instead of jumping from 2 to 5 as I apparently did. ;) It's an easy fix to remove the visited state, so I think we can merge this PR and move the link color discussion into an issue.

PR #399, opened 2015-04-01
Specify file to use for Chosen dependency

Scotchester:
Have to do this to get this hotfix to build, too :crying_cat_face:


cfarm:
Lookin' good :hamster:

PR #402, opened 2015-04-02
Explicitly require jsdom

contolini:
Attempting to placate Travis. :massage:


cfarm:
OMG it worked! 

can you fork that jsdom to owning-a-home instead of your personal account?

contolini:
Let's hold off on implementing this solution. I'm going to do a bit more investigating.

PR #408, opened 2015-04-07
Journey bug fixes

virginiacc:
### Changes
- Update href for journey phase back link
- Display action step numbers using post order attribute
- Don't use webfont for bulleted lists in action steps
- Increase margins before headers in action steps
- Use url_for in phase template so images are displayed locally
- Use gray-5 instead of gray-10 as background color so links are accessible (should probably create a variable for this -- maybe at the CF level)
- Remove 'clarify your goals' link

### Screenshots
screen shot 2015-04-07 at 4 43 18 pm

screen shot 2015-04-07 at 4 43 39 pm

### Review
@cfarm 
@stephanieosan

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -30,7 +30,7 @@
             </div>
 
             <div class="content-l_col content-l_col-1-2">
-              <img src="{{base_url}}static/img/process.png" alt="Illustration of the home-buying process">
+              <img src="{{url_for('static', filename='img/process.png')}}" alt="Illustration of the home-buying process">
           </div>
           </section>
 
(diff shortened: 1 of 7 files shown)


cfarm:
:+1: Looks good!

This is a notice for content editors: Make sure you're adding `<p>` tags around the paragraph text or else the web font and whitespace won't be applied - see example screenshot of why that looks bad

**Before**
screen shot 2015-04-08 at 11 30 38 am


**After**
screen shot 2015-04-08 at 11 32 18 am

Scotchester:
That's interesting, @cfarm. I assumed that our WYSIWYG fields would automatically generate `p` tags, as the main WordPress content field does. If it's not doing that, we can probably figure out how to get it to.

cfarm:
@Scotchester yeah that's what I assumed too! That would be great if it could do that.

stephanieosan:
I second all that! It would be really great if Wordpress was automatically adding p tags. 

I've also noticed that when I manually add p tags in the text editor mode, if I go to edit something in visual mode, then return again to the text editor, the p tags I just manually added get removed! Boo. 

Let me know if that doesn't make sense.

stephanieosan:
Also, @vczosek's changes look great!

PR #409, opened 2015-04-08
add h3 links for journey steps

cfarm:
## Additions
* Adds link to the h3 for each journey step on landing page
* Adds link styling so the h3 is black by default, blue on hover/active/focus

## Review
* @stephanieosan or @virginiacc


virginiacc:
Nice! :+1:

PR #410, opened 2015-04-08
updated behave version, improved debug message, mod for dp amount

OrlandoSoto:
- Updated to behave 1.2.5 for better output log info
- Updated debug messages in the browser tests
- Added some code to ensure our DP amount does not get overwritten by the API and causing our tests to fail

### DIFF
--- a/test/browser_testing/features/pages/rate_checker.py
+++ b/test/browser_testing/features/pages/rate_checker.py
@@ -10,6 +10,7 @@ from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
 
 from pages.base import Base
+from pages.screenshot import Screenshot
 
 # ELEMENT ID'S FOR TEXTBOXES
 DOWN_PAYMENT_AMOUNT_TBOX = "down-payment"  # DOWN PAYMENT AMOUNT TEXTBOX
@@ -55,6 +56,7 @@ class RateChecker(Base):
                                           driver, driver_wait, delay_secs)
         self.logger = logger
         self.driver_wait = driver_wait
+        self.screenshot = Screenshot(self, Base)
 
     # ALERTS
     def get_warning_button_class(self):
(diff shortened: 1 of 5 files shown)


cfarm:
@OrlandoSoto what behave command should I run for this?

OrlandoSoto:
@cfarm run:
- behave -k features/rate_checker_loan_amount.feature 

The output log will show a debug message
- Actual downpayment percent is:

But I just noticed that the down payment percent field does not have a text attribute, so I will change the test to read the value attribute instead

cfarm:
I had one failure for these:

Given I navigate to the "Rate Checker" page                                         # features/steps/steps_navigation.py:22 12.032s
      Assertion Failed: 
      Expected: 'Chart is loaded'
           but: was 'Chart is NOT loaded'

is that expected? I noticed when I manually navigate to the page locally the chart doesn't load right away

cfarm:
I ran this a subsequent time and it did not fail on that step or any other.

OrlandoSoto:
in steps_navigation.py, line 33 I used as a workaround:
- context.base.sleep(1)
which causes the test to wait 1 second for the chart to load.
For now we can probably increase the sleep command to 2 or 3 seconds but we should open up a task to optimize that area, otherwise the smoke tests will take longer to execute

cfarm:
@OrlandoSoto do you want to add this change? If not we can leave it too. I haven't been able to replicate that failure since the first time.

OrlandoSoto:
yes, please. That code will take a snapshot if the county field does not get highlighted. I will probably add more sections where we take a screenshot if the test fails, that way we get additional visibility into the failure.

cfarm:
I meant your change you described but did not commit: "in steps_navigation.py, line 33 I used as a workaround:"

OrlandoSoto:
@cfarm OK I bumped it up to 2 seconds,, I'm looking for a more robust way to do this though

cfarm:
:+1: tests are passing consistently so let's do this

cfarm:
@OrlandoSoto after you fix conflicts that is :)

OrlandoSoto:
@cfarm ok, I fixed it

PR #411, opened 2015-04-08
Update dependencies

contolini:
The acorn browserify bugs (OAH-notes#896) were fixed by its maintainer so here's a new shrinkwrap file that uses the newer version of acorn. Plus grunt-browserify was really old so I updated it.


cfarm:
it works! :tada:

PR #422, opened 2015-04-20
Form explainer fix

virginiacc:
### Changes
- Fix 'closure' typo
- Don't hide form explainer pages until image size has been calculated
- IE8 fixes: assign onload callback before img attr in image resize function, and assign a width to $imageMapWrapper

### Screenshots
screen shot 2015-04-20 at 4 39 00 pm

### Review
@cfarm

### DIFF
--- a/src/form-explainer/closing-disclosure-1.html
+++ b/src/form-explainer/closing-disclosure-1.html
@@ -1,6 +1,6 @@
 {% set data = {
 
-"img": "/owning-a-home/static/img/closing-disclosure-H25B-1.gif",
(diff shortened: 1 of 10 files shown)


cfarm:
This is a grammar thing that I don't think was discussed before but I want to point it out:

'Already have a Closing Disclosure, explore it now' is ungrammatical

Something like 'Already have a Closing Disclosure? Explore it now' makes more sense.

cfarm:
Checked this out locally and in IE8 and it's looking great!

PR #423, opened 2015-04-21
Loan Comparison: add points to rate requests

virginiacc:
### Changes
- Add points to rate requests

### Review
@cfarm


cfarm:
:+1: merging

PR #425, opened 2015-04-22
Journey: temporarily revert to using local phase images

virginiacc:
### Changes
- use local images instead of WP ones for journey phases

### Review
- @cfarm


cfarm:
:+1:

PR #427, opened 2015-04-27
fix trailing slash to match the front end

cfarm:
This should fix current failing browser test on dev site

you can run 

`behave -k features/loan_fha.feature -t=~prod_only`

to test but it will still fail locally because of the URL structure, but should work on dev site.

### DIFF
--- a/test/browser_testing/features/loan_fha.feature
+++ b/test/browser_testing/features/loan_fha.feature
@@ -17,7 +17,7 @@ Scenario Outline: Test Navigational links in the FHA Loan page
 Examples:
   | link_name       	  	     | relative_url											            | page_title 	  |
   | Owning a Home              | /                                            | Owning a Home |
-  | Understand loan options    | /loan-options/  	    								        | Loan Options  |
+  | Understand loan options    | /loan-options  	    								        | Loan Options  |
   | More on mortgage insurance | /loan-options/FHA-loans/#mortgage-insurance  | Loan Options  |
 
 @smoke_testing @loan_options


cfarm:
@OrlandoSoto please review

OrlandoSoto:
this is weird, when I manually navigate to loan options from my mac I get the trailing /

http://build.consumerfinance.gov/owning-a-home/loan-options/
and
http://www.consumerfinance.gov/owning-a-home/loan-options/

cfarm:
me too. ¯\_(ツ)_/¯

OrlandoSoto:
hmm so now if we take out the "/" it fails against build :(

OrlandoSoto:
I have an idea why this build failed with url's don't match
http://zusa.demo.cfpb.gov/view/OAH/job/OaH-build.cf.gov-browser-tests/274/

OrlandoSoto:
Here's my theory:
- The test navigated to http://build.consumerfinance.gov/owning-a-home/loan-options/FHA-loans/
- It clicked on "Understand loan options" which has "href="/owning-a-home/loan-options"
- The failure snapshot shows "Bad gateway - connection error - connection reset by peer"
- The browser address bar was probably stuck on "http://build.consumerfinance.gov/owning-a-home/loan-options'" as directed by the href value so the test failed
- The page never redirected to "http://build.consumerfinance.gov/owning-a-home/loan-options/"

OrlandoSoto:
image

OrlandoSoto:
but then there's also a chance that I'm losing my mind here :)

cfarm:
@OrlandoSoto so this fix is not needed, then?

OrlandoSoto:
@cfarm correct!

PR #432, opened 2015-05-13
Loan comparison results table

cfarm:
This converts the markup for the Loan Comparison outputs to a table instead of many divs. Also includes updated CSS to match the latest design iteration.

## Additions
- New template partial for the results markup to keep files manageable in size

## Removals
- Removed much of the divs and expandables markup from previous version
- Deleted obsolete/duplicate CSS for cf-tables

## Changes

- Table markup for the results
- CSS to style them like the latest design

## Testing

- Check out this branch view locally
- Includes all responsive styles - I kept the same styles as the previous mobile layout, since I didn't have an updated mobile design to work from.

## Notes

- Tooltips aren't functional in this section yet, hence the color

## Review

- @virginiacc 
- @huetingj 

### Small
screen shot 2015-05-13 at 11 24 38 am
screen shot 2015-05-13 at 11 24 23 am

### Medium
screen shot 2015-05-13 at 11 26 41 am
screen shot 2015-05-13 at 11 26 50 am

### Large
screen shot 2015-05-13 at 11 27 32 am
screen shot 2015-05-13 at 11 27 48 am


stephanieosan:
Largely these look really good, @cfarm! 

I think the type hierarchy tweaks @huetingj and I should focus on are the following:
- On desktop, how do we make sure that third (lowest) level looks like a subset of the second level, since we don't have the benefit of an indent, like we do on mobile. Maybe it's as simple as using grey instead of black, we'll explore.
- On desktop, should we bump up the hierarchy of the first level or not.

stephanieosan:
Oh, this is a small thing I just noticed!

On mobile, the rules are between the title and the value, like so:
https://cloud.githubusercontent.com/assets/702526/7614113/b37bbcb8-f962-11e4-988c-e27cb253b854.png
This is probably left over from how the design used to look.

Where as on desktop, the rules separate whole items, like so:
https://cloud.githubusercontent.com/assets/702526/7614170/24a4ff62-f963-11e4-832d-d57b882e2ebe.png

I think we want the mobile rule placement to match the desktop rule placement. Let me know if that doesn't make sense. This is also probably not super-high priority.

cfarm:
I kept the mobile design pretty much the same as how it was before, and only focused on implementing the updated desktop design, since I don't have a new mobile comp. So yep, the border style for mobile is leftover from before.

Rather than capturing those changes here, should we note them in our JIRA task for the mobile outputs design? Or how do y'all want to make those updates? I'd prefer an updated comp so I can submit as a future pull request, just so it's done in one fell swoop.

stephanieosan:
Yes, we'll definitely do a comp. I was mostly writing that as reference for myself, when Jenn and I look at this together tomorrow :)

virginiacc:
This looks great! :+1: I had one question about whether a table can have more than one thead element, but we can look at that after this is merged.

stephanieosan:
Yay, merges!!!

PR #443, opened 2015-05-26
Refactor loan comparison outputs

cfarm:
Moves tables markup from #432 into React components + dynamically displays outputs in the tables

## Additions

- React components for:
 - Output table group
 - Output table
 - Table Head
 - Table Body
 - Head table row
 - Body table row
 - Table cell
 - Static content for mobile output table as placeholder til design handoff
 - Labels added for all outputs that are in the calculations
 - additional properties for the loans that weren't already included in mortgage calculations module

## Changes

- Changed `overall-cost` property to `overall-costs` to match the design copy

## Todo in future PR's

- fix expandables - not working in the react components currently
- mobile outputs
- tooltips logic
- regular outputs logic for the markup for nested results rows - this will fix the styling to match the design
- additional outputs for Closing costs using latest mortgage calculations

## Testing

- Check this out locally to view the outputs and their values in the expandables

## Review

- @virginiacc

## Preview

screen shot 2015-05-26 at 1 36 41 pm
screen shot 2015-05-26 at 1 36 47 pm


src/static/js/modules/loan-comparison/components/loan-output-table-cell.js
```diff
@@ -0,0 +1,28 @@
...
+var common = require('../common');
+var LoanOutput = require('./loan-output');
+
+var LoanOutputCell = React.createClass({
+    displayClassNames: function(loan, prop, type) {
+        var loanID = loan['id'] === 0 ? 'a' : 'b',
+            propResult = prop + '-display-' + loanID,
+            loanScenario = 'lc-result-' + loanID,
+            classes = propResult + ' ' + loanScenario + ' lc-result';
+        if (type === 'primary') {
```
virginiacc: Do we still need the loan-specific class names? If not, we could move these classes (if they're still being used for styling) & the `primary` class to the rows instead

src/static/js/modules/loan-comparison/components/loan-output-table-row.js
```diff
@@ -0,0 +1,70 @@
...
+                            <span className="lc-tooltip" data-toggle="tooltip" role="tooltip" data-original-title="" title=""><span className="cf-icon cf-icon-help-round"></span></span>
+                        </h5>
+                    </th>
+                );
+            }
+        };
+        var loans = this.props.loans.map(function (loan) {
+            return (
+                <LoanOutputCell loan={loan} prop={this.props.prop} resultType={this.props.resultType} />
+            )
```
virginiacc: If we're not using it to generate classes, we could probably create the cells here instead of in a LoanOutputCell component.

src/static/js/modules/loan-comparison/common.js
```diff
@@ -153,7 +153,27 @@ common.errorMessages = {
...
+    'prepaid-expenses': 'Prepaid expenses',
+    'initial-escrow': 'Initial escrow deposit',
+    'monthly-taxes-insurance': 'Monthly taxes and insurance',
+    'monthly-hoa-dues': 'Monthly HOA dues', 
+    'monthly-principal-interest': "Monthly principal and interest",
+    'monthly-mortgage-insurance': 'Monthly mortgage insurance',
+    'loan-term': 'Loan term',
+    'principal-paid': 'Principal paid',
+    'interest-fees-paid': 'Interest and fees paid'
 };
```
virginiacc: We could probably just have entries here where the labels don't match the property names. In the inputs table, I'm generating labels where possible -- we can move that code into a function here for reuse.

src/static/js/modules/loan-comparison/components/loan-output-table-cell.js
cfarm: These are used for styling, since items in column B get additional right padding. I might end up moving the primary class to the rows when I pull in styling for the other rows to make them consistent.

cfarm:
@virginiacc this should be ready to merge! I'll revisit the table cell classes next up.

virginiacc:
:+1:  Looks great!

PR #454, opened 2015-06-02
Fix processor bug

kurtw:
Removes lines that breaks indexing


kurtw:
paging @fna

fna:
works, merging

PR #458, opened 2015-06-03
Linked fields

virginiacc:
### Additions
- Link icons & disabled second inputs for linked fields in scenarios
- Disabled dropdown background image
- `styled-text-input` component to add dollar & percentage signs on down payment & price

### Changes
- Break out `loan-input-table-cell` component from `loan-input-table-row`
- Minor revisions to input components to pass in `disabled` attribute

### Screenshots
screen shot 2015-06-02 at 8 10 12 pm

### Review
@cfarm 
@stephanieosan 
@huetingj


huetingj:
:+1: sweet!

cfarm:
I checked this out locally and I can't edit the down payment inputs when in the down payment scenario. Is this intended?

cfarm:
Finding the same issue in the 'Design your own scenario' interface for down payment inputs

virginiacc:
@cfarm Nope, not intentional. :) Should be fixed now...

src/static/css/module/loan-comparison.less
```diff
@@ -751,11 +747,66 @@
...
+        }
+      }
     }
   
-    tbody tr {
+    tr {
       border: none;
+      
+      &.padded-row {
+        td.label-cell {
```
cfarm: after testing we should revisit the nesting in this section and see if we can reduce it.

src/static/js/modules/loan-comparison/components/loan-input-table-row.js
```diff
@@ -47,66 +12,73 @@ var LoanInputRow = React.createClass({
...
         scenario: React.PropTypes.object // or null
     },
-    handleChange: function (loanId, prop, eventOrVal) {
-        var val;
-        if (typeof eventOrVal === 'object' && eventOrVal) {
-            val = eventOrVal.target.value;
-        } else {
-            val = eventOrVal;
+    
+    generateClassName: function (rowType) {
```
cfarm: I'm going to start using the same naming conventions as you so we keep these functions consistent between components.

cfarm:
Yay it works great now :+1:

PR #459, opened 2015-06-03
Change labels

fna:


amymok:
### Screenshots

I didn't include every possible changes because a lot of them used the same piece of code to generate the changes, so I did some spot check myself without creating all the screenshots, i.e. Header changes to "What to do now".  The same piece of code is used to generate contents for every action steps and pages, therefore, code changes only happened on 1 place, and it will work throughout, and I did spot check just in case anything went wrong.

"During this phase, you will" instead of "During this phase, you should", same scenario as above, I checked all other pages, they all looked good, just included 1 screenshot here:
screen shot 2015-06-08 at 2 07 24 pm

"How to take action" to "What to do now", the scenario I described in the intro paragraph, I spotted check to make sure things look right:
screen shot 2015-06-08 at 2 10 43 pm

"Compare loan packages" to "Compare loan offers" and "Decide and close" to "Get Ready to close":
screen shot 2015-06-08 at 2 12 41 pm

From single quotes to double quotes  on "Hide "xxx" sections".  Same scenario as described, I spot checked to make sure everything looks good:
screen shot 2015-06-08 at 2 19 20 pm

URL from Decide to Close was not changed though.

@stephanieosan

stephanieosan:
These look good!

Just to double-check, the only thing I don't see any screenshots of is the top and bottom nav—did the titles get updated to "Compare loan offers" and "Get ready to close" there as well? If so, let's merge! 

Also, it would be great if someone else would update to the url from decide to close, since Serghei's on vacation.

amymok:
Oops, here they are @stephanieosan:
screen shot 2015-06-08 at 7 52 19 pm
screen shot 2015-06-08 at 7 53 06 pm

amymok:
@stephanieosan I can look into the URLs.  Probably need to be a separate branch and pull request though since I won't be able to push to his branch.

So I think everything is good here, and I will work from a different branch for the URLs.

stephanieosan:
Sounds good! Let's merge!

PR #460, opened 2015-06-04
Downpayment & loan input display updates

virginiacc:
### Additions
- Add flush-left styling to inputs table on desktop

### Changes
- Fix concatenation bug in closing costs calculation by ensuring that `downpayment` is a number
- Move `downpayment-percent` calculation out of downpayment component, since this value needs to be displayed in the mobile loan summary section
- Update `loan-input-table-cell` to accommodate change to dp component, and to break component props generation into several methods
- Add `updateDependencies` method back to loan store to calculate dp/dp-percent
- Refactor loan store's `updateLoan` function slightly so it can also be used by `resetLoans`

### Screenshots

screen shot 2015-06-04 at 12 47 05 pm

### Review
@cfarm 
@huetingj 
@stephanieosan

### DIFF
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -154,10 +154,6 @@
   .webfont-regular();
   position: relative;
   
-  .respond-to-min(@tablet-min, {
-    padding-bottom: unit(@grid_gutter-width / @base-font-size-px, em);
-  });
-  
   .comparison-header {
     .u-hide-on-mobile;
   }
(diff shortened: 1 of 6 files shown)


cfarm:
Styles look good, confirm bug is fixed :+1:

cfarm:
I notice a new bug now (seems to be):

In DP scenario or design your own, all other inputs being equal, when you increase one down payment so it's greater than the other, the interest+fees paid increases and thus overall costs increase. This doesn't seem right. Closing costs should increase, but overall costs shouldn't.

We can troubleshoot that bug in a future branch or perhaps..write some test :dancers:

huetingj:
looks great!

PR #461, opened 2015-06-04
remove maintain folder

fna:


amymok:
Looks good to me (it was hard to tell the maintain folder was actually removed from this branch, but I verified it by looking at Serghei's branch).

@cfarm @virginiacc

cfarm:
:+1:

PR #463, opened 2015-06-08
Display notes

virginiacc:
### Changes
- Scope loan input table styling more specifically to `#loan-input-table` (instead of `table.unstyled`)
- Fix typo in median interest rate calculation (should get median from `totalRates` instead of `rates`)
- Updates to loan input table header styling

### Additions
- Add educational notes & first pass at styling/positioning them on desktop
- New content for input/output tooltips

### Screenshots
screen shot 2015-06-08 at 1 43 05 am

### Review
@cfarm 
@huetingj 
@stephanieosan

### DIFF
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -740,18 +740,23 @@
     }
   }
   
-  table.unstyled {
-    background: none;
-      
-    thead {
-      background: none;
+}
+
+table.unstyled {
+  background: none;
     
-      th {
-        color: @black;
-        background: none;
-      }
-    }  
-  }
+  thead {
+    background: none;
+  
+    th {
+      color: @black;
+      background: none;
+    }
+  } 
+  tbody > tr:nth-child(odd) > th,
+  tbody > tr:nth-child(odd) > td {
+      background: none;
+  } 
 }
 
 .scenario-alternates {
@@ -834,14 +839,14 @@
 }
 
 
-#loan-input-table {
+.loan-input-table {
   width: 100%;
   position: relative;
   
   select {
     font-size: unit(14px / @base-font-size-px, em);
   }
-
+  
   th,
   td {
       width: 33%;
@@ -854,11 +859,7 @@
       
       &.input-1 {
         padding-left: unit(8px / @base-font-size-px, em);
-      }
-
-      tbody > tr:nth-child(odd) > & {
-          background: none;
-      }
+      }      
       
       &.link {
         width: 20px;
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -157,10 +157,6 @@
   .comparison-header {
     .u-hide-on-mobile;
   }
-  
-  .help-text {
-    display: none;
-  }
 
   input[type="text"] {
     width: 100%;
@@ -1039,6 +1035,15 @@
   }
 }
 
+/* Tooltips */
 .tooltip {
   .webfont-regular();
+  .tooltip-inner {
+    max-width: 325px;
+  }
+}
+
+
+.help-text {
+  display: none;
 }
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -746,7 +746,6 @@
   
   table.unstyled {
     background: none;
-    width: 100%;
       
     thead {
       background: none;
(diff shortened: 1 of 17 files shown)


stephanieosan:
These look nice!

src/static/css/module/loan-comparison.less
```diff
@@ -942,3 +834,216 @@
 }
 
 
+#loan-input-table {
```
cfarm: can this be a class instead of an ID?

src/static/js/modules/loan-comparison/components/app.js
```diff
@@ -30,16 +33,52 @@ var App = React.createClass({
...
             selector: '[data-toggle="tooltip"]',
             'placement': 'bottom', 
+            container: 'body',
             title: function getTooltipTitle(){
                 return $(this).attr('title') || $(this).next('.help-text').html() || 'Tooltip information.';
             }
         });
         $('.expandable').expandable();
+        
+        var animating, scenario = this.state.scenario;
```
cfarm: can you indent scenario here for readability

src/static/js/modules/loan-comparison/components/loan-output-table-row.js
```diff
@@ -41,15 +48,20 @@ var LoanOutputRow = React.createClass({
...
+            className += note ? ' highlight' : '';
+        }
         
-        className += educationalNote ? ' highlight' : '';
-
         return (
             <tr className={className}>
                 {headingType(prop, resultType, label, scenario)}
                 {loans}
+                <td className="educational-note">{noteHtml}</td>
```
cfarm: might make sense to move this to the `LoanOutputCell` component at some point.

src/static/css/module/loan-comparison.less
virginiacc: Sure.

src/static/js/modules/loan-comparison/components/app.js
virginiacc: Done!

src/static/js/modules/loan-comparison/common.js
```diff
@@ -21,8 +21,8 @@ common.scenarios = [
         label: 'down payment',
         intro: "Your down payment amount affects all aspects of your costs. Putting down less up front can be a good option for home buyers without a lot of cash on hand, but you’ll have higher monthly payments and pay more in interest and fees. This tool will help you get a sense of how much the difference in costs is likely to be, so you can make tradeoffs.",
         loanProps: [
-            {downpayment: 40000}, 
-            {downpayment: 20000}
+            {'downpayment-percent': 20},
+            {'downpayment-percent': 10}
         ],
         sharedProps: {},
```
cfarm: is `sharedProps` used anywhere? I couldn't find it, and it seems like we're defining `independentInputs` instead for the linked functionality

cfarm:
Looks great and I'm so impressed by the genius of the position-notes function :) 

This isn't an issue for the prototype and the down payment scenario, but I can see this possibly coming up in a future scenario: 

If all 3 output headings have notes, then there's some funky overlap that can happen:

1. Load scenario with 3 output heading notes
screen shot 2015-06-09 at 4 22 15 pm

2. Click to expand either Monthly Payment or the Overall Costs expandable. When you do this, the notes adjust to position at the top of the output table - whether or not it's overlapped with a note in the table above
screen shot 2015-06-09 at 4 24 03 pm


3. Closing the expandable, the repositioned note persists at the top of the table instead of below other notes

screen shot 2015-06-09 at 4 17 27 pm

If it's an extensive fix, we can create a new story for it for the backlog and go ahead and merge, so I'l defer to @virginiacc on that decision - the only other notes i had were super minor.

virginiacc:
@cfarm Oh, good catch! Yeah, that's a side effect of skipping notes above the animating expandable without storing their positions. I'll push a fix for it. 

There probably still will be an issue with two adjacent expandables that both have header notes -- if you close the first expandable, the second expandable's header note will temporarily overlap the first one's until animating finishes. This will be a more involved fix, so I'll hold off on implementing it for the moment.

stephanieosan:
@huetingj and I talked about this potential spacing issue a while ago. 

One thought we had was this: what if we just had a single, summary note that explained all three roll up sections at once? And maybe when you expand any sections, it could stay fixed to the top of the "Cash to close" section? Or it could disappear and be replaced by the notes within the expanded sections. This would require a content re-write from Megan.

But, totally open to exploring what solutions might work best with how the code is written.

huetingj:
+1 to having one summary note for all three sections

stephanieosan:
Also, agreed to merging it as-is and coming back to fix the roll-up issue with another story.

virginiacc:
@stephanieosan @huetingj I also like the idea of a single summary note. From an implementation standpoint, it would definitely simplify positioning an expandable's notes if we only show them when it's open.

PR #466, opened 2015-06-08
subtract down payment from closing costs for overall-loan-cost object

cfarm:
The overall-loan-cost module takes properties for closing costs and down payment separately, therefore down payment should not be included in the closing-costs value. This bug causes the overall costs to increase if your down payment is bigger, which is wrong.

## Changes

- calculations object for the overall-loan-cost module now subtracts down payment from closing-costs since *our* closing-costs property includes dp, but the module needs them passed separately

## Testing

- We need to write tests to make sure this is working as expected, on the docket for 6/9

## Review

do not review until you see tests here


virginiacc:
@cfarm, this looks good -- the results seem pretty accurate when compared to our test scenarios (some of the other costs are different, so they're not an exact match). Maybe we should pull it in now for user testing?

cfarm:
That's fine with me, :shipit: 

I was thinking the test requirement would force us to get on that but that was before we hit any obstacles.

PR #472, opened 2015-06-11
Loan comparison illustration

stephanieosan:
And CSS fixes to the page title.

## Additions
- Added @huetingj's latest illustration for testing (goodby, kitten!)

## Changes
- Made title of the tool an H1 instead of H3
- Changed the css class of the <Owning a Home link above the tool's title to match the class that the same feature on the rate checker uses (this bumped up the link to 22px and added the right amount of space between the link and the header)
- removed 30px of space from top of the scenario container (this brings the back to owning a home link closer to the page title)

## Removed
- stray periods in titles

## Check it out!
screen shot 2015-06-11 at 4 32 23 pm

## Notes
- PSYCH, this is fixed! I would still like to get "use our presets" and "design your own scenarios" lining up, but I haven't figured out how yet, because they both have the same class so I'm not sure why they're not behaving in the same way.
- ALSO FIXED! The illustration is pretty big. I want advice on how I should make it smaller. I want it to match up with the ratechecker at 200px, which is the size the image file is, too. But I think it's showing up closer to 250px. Is it possible that it's stretching to be the size of the grid it's hanging out in?

## Review
@huetingj
@cfarm or @virginiacc


cfarm:
cool beansssss

@huetingj your illustration is beautiful, but is there anyway we can work the cat back in? maybe as one of the icons??

cfarm:
cool beansssss

@huetingj your illustration is beautiful, but is there anyway we can work the cat back in? maybe as one of the icons??

cfarm:
cool beansssss

@huetingj your illustration is beautiful, but is there anyway we can work the cat back in? maybe as one of the icons??

cfarm:
cool beansssss

@huetingj your illustration is beautiful, but is there anyway we can work the cat back in? maybe as one of the icons? thx in advance

cfarm:
@huetingj here is a mockup, what do you think ?
loan-comparison-kitten

Scotchester:
Done. Don't see why we need Jenn here. :shipit:

cfarm:
@stephanieosan code looks good and looks like you added changes for the stuff you said you were stuck on in Notes, :+1:

huetingj:
so much better!!! :heart_eyes_cat: :heart_eyes_cat: :heart_eyes_cat:

stephanieosan:
Yesss, I added the code @virginiacc helped with!

stephanieosan:
Am I allowed to hit the merge button myself, since everyone else reviewed? Am I not allowed to merge until I replace the image with the cat-loan version? How does this work?! I'm too excited.

PR #473, opened 2015-06-11
Loan store bug fix & validation updates/prelim tests

virginiacc:
### Changes
- fix `loanUpdate` null value bug
- change `styledTextInput` to `styledNumericInput` to reflect its purpose in styling `downpayment` & `price` inputs
- enforce numeric input in `styledNumericInput` fields
- change `updateLoanState` to `validateLoan` and separate out each property's validation
- change `ErrorMessage` component to more general `Message`
- revise error message display and error class addition in `LoanInputTableCell` component

### Additions
- add `loan_store_test` file and a few sample tests for loan store validation

### Notes
- temp. changed `oah-config` path in `mortgages.js` and `rates.js` to get tests running -- this needs to be changed back


src/static/js/modules/loan-comparison/components/input-text.js
```diff
@@ -11,17 +11,25 @@ var DebouncedTextInput = React.createClass({
             React.PropTypes.bool
         ])
     },
+    getDefaultProps: function() {
```
cfarm: Neato, https://facebook.github.io/react/docs/reusable-components.html#default-prop-values

virginiacc:
@cfarm, @amok I'm going to close this PR and submit another one without the tests so we can get these changes in while we resolve the issue with testing dependencies.

PR #474, opened 2015-06-12
Fix loan store bug

virginiacc:
### Changes
- fix `loanUpdate` null value bug
- change `styledTextInput` to `styledNumericInput` to reflect its purpose in styling `downpayment` & `price` inputs
- enforce numeric input in `styledNumericInput` fields
- change `updateLoanState` to `validateLoan` and separate out each property's validation
- change `ErrorMessage` component to more general `Message`
- revise error message display and error class addition in `LoanInputTableCell` component

### Notes
- removed tests so these changes can be merged while we resolve testing dependency issues

### Review
@cfarm 
@amymok


src/static/js/modules/loan-comparison/components/styled-numeric-input.js
```diff
@@ -1,12 +1,20 @@
 var React = require('react');
 var common = require('../common');
 var TextInput = require('./input-text');
+var nonNumericRegex = /[^0-9.]+/g;
 
-// adds a unit (dollar or percent sign) to input
-var StyledTextInput = React.createClass({
+// Enforce numeric input
+var numericValidator = function (val) {
```
cfarm: We should discuss with designers how this should work post-user testing - I would prefer an error message if you try to type with text to avoid confusion

cfarm:
Grood

PR #478, opened 2015-06-16
Downpayment modes

virginiacc:
### Changes:
- adds `downpaymentConstant` to reflect last updated down payment field & determine which down payment field is recalculated when price changes
- refactors `updateDependencies` to work with `downpaymentConstant`

### Notes:
- How this works:
    - One of the two down payment fields -- `downpayment` & `downpayment-percent` -- is constant when price changes. It stays the same, and the other field gets calculated from the constant & the changed price.
    - The app always starts with `downpayment-percent` as constant.
    - When the user makes a change to one of the down payment fields, the down payment constant changes globally -- the most recently updated property becomes the one that is constant for both loans.
    - `downpaymentConstant` is carried over when switching between scenarios -- EXCEPT when entering the `downpayment` scenario, which always starts with `downpayment-percent` as constant.

### Review
@cfarm @amymok

### DIFF
--- a/src/static/js/modules/loan-comparison/stores/loan-store.js
+++ b/src/static/js/modules/loan-comparison/stores/loan-store.js
@@ -49,6 +49,8 @@ var validators;
 
 var _loans = [];
 
+var downpaymentPercentMode = true;
+
 function init () {
     resetLoans(true);
 }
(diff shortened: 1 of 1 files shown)


stephanieosan:
Just to note, we want this reviewed and merged in before testing tomorrow, if possible.

cfarm:
:+1:

PR #479, opened 2015-06-16
Fix Steps Numbers in the Journey

amymok:
### Changes
* Use loop index so we don't depend on the correctness of the step order if an item has been removed

### Review
@cfarm @virginiacc

### DIFF
--- a/src/process/_process-step.html
+++ b/src/process/_process-step.html
@@ -93,7 +93,7 @@
                     <div class="expandable expandable__padded">
                       <button class="expandable_header expandable_target">
                         <span class="expandable_header-left expandable_label">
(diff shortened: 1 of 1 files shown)


cfarm:
@amymok do you have any screenshots? How can we test this locally?

amymok:
@cfarm  I can do some screenshot but nothing really changed visually so I didn't do it.

The way I do to test it is to go into wp on build, change the order number for the last number, i.e. in Prepare, step 8 will be changed to 9 or above.

Without the changes from this branch, your local version will have 9 instead.  But if this branch has been applied, the local version will have 8.

Remember to reindex to see changes happening.

cfarm:
:+1: Tested and it works.

PR #484, opened 2015-06-18
Test journey links

fna:
Checking that the main /process/XXX links are working and all links of the same form on each of those pages are working too.

### DIFF
--- a/test/browser_testing/features/environment.py
+++ b/test/browser_testing/features/environment.py
@@ -10,6 +10,7 @@ from selenium import webdriver
 from pages.screenshot import Screenshot
 from pages.base import Base
 from pages.home import Home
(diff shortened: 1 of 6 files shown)


fna:
Now on local box tests will only check "local" links ( localhost/owning-a-home/process/XXX ), because there are links to /es, /hmda, etc which will return 404, so it would be a false positive. In other environments it will check all links. 

A status code of 2XX and 3XX is considered good. 3XX codes I need to not fail external services, like youtube which redirects to www.youtube.

cfarm:
@amymok can you review this?

amymok:
Tests ran locally with no errors.  Good to go!

stephanieosan:
yay! merge?

PR #489, opened 2015-06-24
More form explainer copy changes

stephanieosan:
These came from Harris.

~~Don't review/merge this yet. Might as well wait until we get the updated links from @mthibos.~~


stephanieosan:
The word from @mthibos is those links aren't coming for a while, and we should add that portion of the update to the backlog. So, I think we can go ahead with reviewing and merging these. 

### Changes
* Content updates from Harris

### Screenshots

### Review
@cfarm or @virginiacc
Harris will review actual content on build once its merged.

virginiacc:
Looks good!

PR #490, opened 2015-06-24
Automatic Testing for Downpayment Scenario & Fix Zombie Downpayment Percentage

amymok:
### Additions
* Unit/Integration tests for in loan-store, specifically on the downpayment scenario

### Changes
* Refactored loan store to make it easier to test
* Fix the issue that oah-config cannot be found
* Fix the zombie downpayment percentage issue.  Once dp % is deleted, it will not reappear now

### Reviews
@cfarm @virginiacc


virginiacc:
Looks great! :+1:

PR #493, opened 2015-06-25
Form explainer resize

virginiacc:
### Changes
- show tab titles on small screens
- refactor form-explainer.js to make code more reusable

### Additions
- resize form images on window resize

### Screenshots
Tab titles on small screen:
screen shot 2015-06-25 at 4 14 39 pm

### Review
@cfarm @amymok @stephanieosan


stephanieosan:
@virginiacc Is that screenshot from a small screen? If so, it looks great! If not, could you post mobile screenshots as well?

virginiacc:
@stephanieosan Oh, yeah, sorry about that, I should have added a label. It's impossible to tell that's a screenshot of the mobile layout when the image is resized. :D

src/static/css/module/tab.less
```diff
@@ -99,7 +99,7 @@
         font-size: 1em;
       }
       .tab-label {
-        display: none;
+        //display: none;
```
cfarm: I think we want to change this to a media query with a max-width of 332px, not just make it visible. At 320px (standard iphoneish) it will not fit and will wrap to a new line.

cfarm:
One CSS fix needed, when that's done this can merge.

PR #494, opened 2015-06-26
Form explainer accessibility

cfarm:
This improves keyboard navigation of the form explainer pages using tabindex and a lil JavaScript.

## Additions

- tabindex attributes for interactive elements including the tabs, the expandables, and expandable content such as links and paragraphs
- JavaScript to manage focus and active selected styles
- add JS to move focus to `#primary-content` div when the skip nav link is clicked

## Changes

- cache `$(this)` for hover stuff
- move skip nav link to nemo header partial
- add nemo header js to form explainer module so the nav works :smirk_cat: 
- properly indent nested nemo divs

## Removals

- removed JS for the obsolete 'all' category

## Testing

1. check out locally
2. hit CMD+F5 to turn on Voiceover
3. tab through the links of the page with the keyboard, hitting shift+tab to go backwards. Each piece of relevant content from the expandables text should be read aloud, and the corresponding overlay should receive the active styling

## Review

- @virginiacc


src/static/js/modules/form-explainer.js
```diff
@@ -286,7 +274,18 @@ $(document).ready(function(){
...
+
+  });
+
+  // Add tabindex to links inside of expandables, so they are tabbable in the correct order.
+  $WRAPPER.on( 'focus', '.expandable__form-explainer .expandable_target', function (event) {
+    var $this = $(this),
+        $content = $this.siblings('.expandable_content'),
+        $tabindex = $content.attr('tabindex');
+    $content.find('a').attr('tabindex', $tabindex);
   });
```
virginiacc: Could we iterate through all the expandables and add tabindex values to their links on page load?

virginiacc:
@cfarm I tried this with voiceover in Chrome & tabbing within the form works great! The only potential issue I see is that the form tab order is preempting the normal tab order, so tabbing through the page you hit the nav links at the top after the form. Should the form be within the usual tab flow, or do we want to start with it?

cfarm:
@virginiacc good catch, I will fix the tab order so it fits into the rest of the page flow instead of preempting it.

cfarm:
@virginiacc updates made and this is ready for your review!

virginiacc:
This looks great -- tabbing & skip link all working for me on Chrome! :+1:

PR #508, opened 2015-07-10
fix third party fees total in outputs

cfarm:
Fixes bug where the third party fees sum was NaN (and appearing as "0" on the front end).

## Changes

- Fixed `calculatedPropertiesBasedOnIR` array order 

## Testing

- check the loan object values on page load to see that `third-party-fees` value is now 3000 instead of NaN.

## Review

- @virginiacc 

## Preview

screen shot 2015-07-10 at 11 14 27 am

## Notes

- I want to write a test for this. This is the second time the order of operations has broken a result (and we haven't caught it through manual or automated testing).


cfarm:
@virginiacc can you review this? I'm going to add the test in a separate branch.

virginiacc:
Confirmed this is fixed. :+1:

PR #510, opened 2015-07-13
Loan comparison print button

virginiacc:
### Additions
- Make LC print button functional
- Add polyfills so React works in IE

### Review
@cfarm


amymok:
@virginiacc are we just adding print capability in this PR without worrying whether the printout looks good or not?

Because I don't think the printout looks very good...

virginiacc:
@amymok Yeah, it could definitely be improved. :) This PR is just to bring up the print dialog when the print button is clicked -- there's a later task to add css to style the printout.

amymok:
If that's the case, this looks good to me!

PR #511, opened 2015-07-13
Loan comparison: auto api calls -- do not merge

virginiacc:
### Changes
- Automatically call api for new rates when loan is changed

### Notes
- Do not merge -- this is just for testing purposes
- This is a quick test, so it doesn't have a loading state for the outputs. The rate input still switches out for a loading icon while the api is being queried.

### Testing
- Check out locally.
- Change a value on a loan. Wait for the interest rate dropdown & outputs to automatically update with new data from the api. 

### Review
@cfarm @stephanieosan @sonnakim

### DIFF
--- a/src/static/js/modules/loan-comparison/stores/loan-store.js
+++ b/src/static/js/modules/loan-comparison/stores/loan-store.js
@@ -99,7 +99,6 @@ var LoanStore = assign({}, EventEmitter.prototype, {
         this.validateLoan(loan);
         this.updateLoanCalculatedProperties(loan, rateChange);
         if (!rateChange) {
-            console.log('not rate change');
             this.fetchLoanData(id);
         }
     },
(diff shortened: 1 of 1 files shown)


stephanieosan:
Okay, now is maybe the time for someone to teach me how to pull down something locally :)

cfarm:
Github will tell you! Click that "command line" link under this comment next to the Merge button:

https://www.dropbox.com/s/xvva5c982d5mtqe/Screenshot%202015-07-13%2021.37.32.png?dl=0

Follow step 1 to checkout another person's changes. and ignore step 2.

stephanieosan:
Wahoo! I followed the instructions, ran grunt and sheer, and was able to see the auto-updating locally. 
BUT, for some reason the data wasn't showing up:

screen shot 2015-07-14 at 10 30 56 am

It was all zeros all the time. I'd like to see how quickly the numbers update, too. How can I fix this?

virginiacc:
@stephanieosan I'm not sure why those aren't updating -- do you see an error in the javascript console in Chrome? Do the outputs update if you change the rate in the dropdown? I wonder if the mortgage insurance api call was failing, since the rates dropdown updates if the rates call completes, but the calculations only run when both calls succeed. I added a line to run the calculations when the rates data is received, so we can see if that works for you locally...

The outputs should change pretty much immediately after the interest rates are updated, though, so the switch from the rates loading state to the dropdown is a good indicator of the time it would take to see new data from an api call.

virginiacc:
This is ready to merge except that it's breaking tests in this branch. The tests are changing in the jumbo loan branch, so I'll implement this again once that code is merged.

PR #512, opened 2015-07-14
check that hash value consists of letters, digits and - only

fna:


stephanieosan:
Who else needs to review this before it's ready to merge?

virginiacc:
Looks good! :+1:

PR #515, opened 2015-07-15
Form explainer 'new' tag

cfarm:
Add 'new' tag element to the form explainer content boxes.

## Additions

- Logic for new tags
- 'new' property on loan estimate's first content box - as example til we have the actual content that needs these.

## Removals

- alert tab styles - we removed the alert tabs from the design so these are no longer needed.

## Changes

- refactored the conditionals for the term (aka content box) so that there is less duplicate stuff

## Testing

- go to [http://localhost:7000/loan-estimate/](http://localhost:7000/loan-estimate/) to see the New tag output on the first content box.
- CMD+F5 to use voiceover and tab through the page to hear the tag read aloud (it's after the content title, to give the user context).

## Review

- @sonnakim 
- @virginiacc 

### Small screen
screen shot 2015-07-15 at 1 44 48 pm

### Large screen
screen shot 2015-07-15 at 1 44 56 pm

### DIFF
--- a/src/loan-estimate/loan-estimate-1.html
+++ b/src/loan-estimate/loan-estimate-1.html
@@ -11,7 +11,8 @@
         "left": "16.17%",
         "top": "15.02%",
         "width": "21.56%",
-        "height": "4.50%"
+        "height": "4.50%",
+        "new": true
     },
     {
         "term": "Check loan terms, purpose, product, and type",
--- a/src/loan-estimate/loan-estimate-1.html
+++ b/src/loan-estimate/loan-estimate-1.html
@@ -12,7 +12,7 @@
         "top": "15.02%",
         "width": "21.56%",
         "height": "4.50%",
-        "new": true
+        "tag": "new"
     },
     {
         "term": "Check loan terms, purpose, product, and type",
--- a/src/_layouts/form-explainer.html
+++ b/src/_layouts/form-explainer.html
@@ -131,64 +131,53 @@
 
 {% macro term( item, index, page_num ) %}
 {#- Some terms are not expandable -#}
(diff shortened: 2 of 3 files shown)


sonnakim:
@cfarm Is that Avenir regular weight italic, 11 pt?

cfarm:
"AvenirNextLTW01-Italic", Arial, sans-serif (i used the existing ['micro-copy' style](https://cfpb.github.io/cf-typography/docs/) from Capital Framework)

sonnakim:
Cool, just wanted to double-check—the web font as rendered looks slightly thicker than it did in the PSD. 

I think this is good to go :+1:

src/loan-estimate/loan-estimate-1.html
```diff
@@ -11,7 +11,8 @@
         "left": "16.17%",
         "top": "15.02%",
         "width": "21.56%",
-        "height": "4.50%"
+        "height": "4.50%",
+        "new": true
```
virginiacc: What if we called this property something like 'descriptionTag' and made 'new' its value? That would make it possible for us to apply different tags in future, if we ever wanted to do so...

virginiacc:
@cfarm Looks great! :+1: I had one minor suggestion about making the tagging reusable, but I think this is ready to go when the content is done.

src/loan-estimate/loan-estimate-1.html
cfarm: good idea! I'll add it.

cfarm:
@virginiacc I made the change you suggested.

cfarm:
No longer needed! @!!!!

PR #519, opened 2015-07-16
Loan comparison: jumbo logic & tests

virginiacc:
### Additions
- Adds jumbo loan check & jumbo messaging in the UI
- Fetches county data & displays in UI
- Unit tests for jumbo logic & county functionality
### Changes
- Refactors some of the loan store code around updating a loan 
- Adapts loan store unit tests to code changes
### Screenshots

screen shot 2015-07-13 at 11 28 47 am

screen shot 2015-07-13 at 11 28 56 am

screen shot 2015-07-16 at 7 23 13 pm
### Review

@cfarm @amymok @stephanieosan @mthibos
### Notes
- Here are the assumptions I made about the jumbo logic, mostly based on the rate checker implementation:
  - **Loan type options in jumbo mode:** When you are in jumbo mode, you can select other loan type options. You will then see jumbo messaging for that type, if you are still in jumbo mode. The type that you are seeing the error for will be unselectable in the loan type dropdown.
    screen shot 2015-07-16 at 7 24 14 pm
  
  to:
  
  screen shot 2015-07-16 at 7 24 23 pm
  - **Previous loan type:** The app stores the last non-jumbo loan type you selected for a loan. If you exit jumbo mode by decreasing your loan amount, you will be switched into that stored loan type. If you change states/counties, the jumbo check will be run for that stored loan type.
  
  screen shot 2015-07-16 at 7 26 31 pm
    to
  
  screen shot 2015-07-16 at 7 26 42 pm
  - **Switching states in jumbo mode:** If you switch states while in jumbo mode, the app continues showing the current jumbo loan type, along with a message on the county field that refers to the previous loan type. Alternatively, we could switch the loan type dropdown back to the previous loan type as well
  
  screen shot 2015-07-16 at 7 29 06 pm
  - **Persistent jumbo error message:** The jumbo state highlight/messaging shows as long as you are in jumbo mode. If you are in jumbo mode & get an ARM term error when switching to an adjustable rate mortgage,  the type dropdown continues to be highlighted, but only the ARM error message is displayed. Alternatively, I could show both messages, or only show the the jumbo state when the user changes one of the inputs that affects it.
    screen shot 2015-07-16 at 7 30 11 pm
  - **County dropdown:** The county dropdown is always shown, even if you are not in jumbo mode, but it defaults to a  message asking the user to select a county.
    screen shot 2015-07-16 at 7 34 10 pm


src/static/js/modules/loan-comparison/api.js
```diff
@@ -1,10 +1,11 @@
 var $ = jQuery = require('jquery');
 var fetchRates = require('../rates');
 var fetchMortgageInsurance = require('../mortgage-insurance');
+var config = require('../../../../../config/config.json');
 
 var api = {};
 
-api.fetchCounties = function (appState) {
+api.fetchCountyData = function (appState) {
     //dropdown('input-county').hideLoadingAnimation();
```
cfarm: can this be removed or otherwise noted how it will be used in future?

src/static/js/modules/loan-comparison/common.js
```diff
@@ -239,17 +240,32 @@ common.options = {
...
+common.jumboTypes = {
+    'agency': {val: 'agency', label: 'Conforming jumbo'}, 
+    'jumbo': {val:'jumbo', label: 'Jumbo (non-conforming)'}, 
+    'fha-hb': {val:'fha-hb', label: 'FHA high-balance'}, 
+    'va-hb': {val:'va-hb', label: 'VA high-balance'}
+};
+
+common.norms = ['conf', 'fha', 'va'];
+
+common.jumboMessages = {
```
cfarm: this can be added to the `common.errorMessages` object

src/static/js/modules/loan-comparison/components/input-select.js
```diff
@@ -9,24 +9,32 @@ var SelectInput = React.createClass({
             React.PropTypes.number,
             React.PropTypes.bool
         ]),
-        title: React.PropTypes.string,
-        disabledItemCheck: React.PropTypes.func
+        //title: React.PropTypes.string,
```
cfarm: if you're planning to use these later, let's add a @todo comment

src/static/js/modules/loan-comparison/components/input-select.js
```diff
@@ -9,24 +9,32 @@ var SelectInput = React.createClass({
...
+            }
         };
     },
     generateOptions: function () {
+        var config = this.props.config;
+        var val = config.valProp;
+        var label = config.labelProp;
+    
         if (this.props.items) {
             var opts = this.props.items.map(function (opt) {
```
cfarm: save `this.props.items` as a variable

cfarm:
This is a note for future work: This PR is TOO BIG! We really need to be able to review PR's in digestible chunks and this one is too large for that. I vote for PR's that consist of no more than 2 days' work in the future.

src/static/js/modules/loan-comparison/stores/loan-store.js
```diff
@@ -37,34 +37,57 @@ var LoanStore = assign({}, EventEmitter.prototype, {
...
-    resetLoan: function (id, scenario) {
+    setupLoanData: function (id, loan, scenario) {
+        var defaultData = common.defaultLoanData;
         var scenarioData = scenario ? scenario.loanProps[id] : {};        
-        var existingData = this._loans[id];
-        
-        if (scenario && id > 0) {
-            existingData = common.omit(this._loans[0], 'id', 'rate-request', 'mtg-ins-request');
-        }
+        var existingData = scenario && id > 0
```
cfarm: so `existingData` only applies to scenarios, correcT?

src/static/js/modules/loan-comparison/api.js
virginiacc: Yeah, that's not being used. I'll remove it.

src/static/js/modules/loan-comparison/components/input-select.js
virginiacc: Good point! Will do.

stephanieosan:
Okay, I've read through all of @virginiacc's notes. Here are my responses.

* **Loan type options in jumbo mode** Yes, good!
* **Previous loan type** Yes, good!
* **Switching states in jumbo mode** 
 * I think it should stay as Jumbo mode, not switch back to previous loan type—that's what we do in explore rates—but the error message is different in that case. See screenshot I just took there: 
screen shot 2015-07-20 at 5 17 38 pm
To get that screenshot, I triggered Jumbo mode in an Alabama county and then switched over to Maine.
* **Persistent jumbo error message** 
 * Yeah, this is wonky. And I don't think Explore rates solves for it. I think we need @mthibos on this one. I'm no expert, but I think Jumbo ARMs don't exist. What we may want to do is, if you select Adjustable while in Jumbo mode, the tool would then bounce you back to fixed and replace the Jumbo message with this message and also highlight the rate type dropdown (in addition to Jumbo being highlighted). What do you think @mthibos?
screen shot 2015-07-20 at 5 22 52 pm
* **County drop down** Yes. 
 * I think this is persistently visible in the Loan comparison because we need the county to calculate some of the outputs? Is that true, @mthibos?

src/static/js/modules/loan-comparison/stores/loan-store.js
virginiacc: There would be existingData when transitioning into custom mode too. The main difference is that in scenarios, the first loan's data serves as existingData for loan 2 as well.

This whole function should go away when we remove the prepop scenarios, though. :)

virginiacc:
@stephanieosan 
Re: **switching states in jumbo mode** -- Yes, that county error messaging makes more sense. I'll implement it.

src/static/js/modules/loan-comparison/common.js
virginiacc: Yes! And @stephanieosan suggested we use the county error message from rate checker, so that will simplify the jumbo messaging.

virginiacc:
@stephanieosan Here's the 'needs county' state with the more general error message:

screen shot 2015-07-20 at 6 07 08 pm

When we switch to automatically fetching interest rates, would we gray out the outputs in this scenario until the user selects a county and we run the jumbo check?

stephanieosan:
Yeah, good catch. I think we'd have to.

mthibos:
Ok!!  Thanks @stephanieosan for getting started with this.

*Counties when switching states*
Because we use county all the time in LC, compared to only in jumbo mode for RC (and yes that's because we need it for other data points not just jumbo rates), I think we can make the error message shorter. 

Instead of "Please enter your county so we can check..." I think it can just be "Please enter your county."  And that error message should probably be triggered whenever the person switches states in the middle of things, regardless of whether we are in jumbo mode or not.  We wouldn't want someone to be stuck with the "A" county without realizing it.  And yes, I think we would grey out the outputs both during the error message until the user puts in a county.

mthibos:
**Jumbos and ARMs**

These can co-exist.  See, eg, this is how we do it on rate checker. 

image

I think the concern was what if they choose an ARM that doesn't exist.  Like an FHA arm or a 15 year ARM.  On rate checker, we display both alerts.  Is that not an option here?

image

[User flow to get this situation:  change state to california > change loan term to 15 years > change DP to 20% > change HP to $600k > change rate type to adjustable]

mthibos:
**Stored loan type.**

The logic we implemented in Rate Checker was actually a bit simpler (I think).  We did not store the last  non-jumbo selected loan type.  Instead, we went with the low-balance equivalent of whatever was currently in the drop down.

That logic is documented here:  GHE/OAH/OAH-notes/issues/621#issuecomment-32613

We thought that was most straightforward. I'll explain more later, but wanted to go ahead and post this now.

stephanieosan:
:+1: I'm cool with whatever @mthibos thinks is best on these. :smile:

src/static/js/modules/loan-comparison/stores/loan-store.js
```diff
@@ -363,17 +460,163 @@ var LoanStore = assign({}, EventEmitter.prototype, {
...
+            // run validations based on new loan type
+            if (newType) {
+                LoanStore.updateLoan(loan, 'loan-type', newType);
+            }
+        } else if (jumboTest.needCounty) {
+            // If the test fails because we need county data we don't have,
+            // update the error messaging & start fetching data if fetch not in progress.
+            loan['need-county'] = true;
+            loan['errors']['county'] = common.jumboMessages[loan['previous-type']];
+            if (common.jumboTypes[loanType]) {
```
cfarm: please add a comment here about why you're keeping this conditional

src/static/js/modules/loan-comparison/stores/loan-store.js
```diff
@@ -363,17 +460,163 @@ var LoanStore = assign({}, EventEmitter.prototype, {
...
+    updateLoanCounties: function (loan, data, initialRequest) {
+        var dict = {};
+        
+        if ($.isArray(data)) {
+            $.each(data, function (num, i) {
+                dict[i['complete_fips']] = i;
+            });
+            loan['county-dict'] = dict;
+            loan['counties'] = data;
+            //if (initialRequest) {
```
cfarm: can this commented out block be removed? or otherwise documented

cfarm:
Special request: I reviewed the changes so far, (except for the tests, they pass and have great coverage so that's good enough for me) can we make any design/logic updates in future PR's so they're easier to review? If so I can go ahead and merge this today.

mthibos:
@cfarm I think that is a great idea.  It is also much easier for me to test live on build.

PR #520, opened 2015-07-17
FE browser tests

fna:
1. Don't test pagination
2. Highlights tests randomly fail/succeed, don't know what's up

### DIFF
--- a/test/browser_testing/features/closing_disclosure.feature
+++ b/test/browser_testing/features/closing_disclosure.feature
@@ -34,7 +34,7 @@ Scenario Outline: Test that resizing window size changes image size too
 
 Examples:
   | tab_name    | css_selector          |
-  | Checklist   | div.illustration img  |
+  | Checklist   | img.image-map_image   |
   | Definitions | img.image-map_image   |
 
 
--- a/test/browser_testing/features/closing_disclosure.feature
+++ b/test/browser_testing/features/closing_disclosure.feature
@@ -1,0 +1,61 @@
+Feature: Test the default values in the Clsoing Disclosure page
+  As a first time visitor to the Owning a Home page
+  I want to have content loaded
+  So that I can get clever and conquer the world
+
+Background:
+   Given I navigate to the "Closing Disclosure" page
+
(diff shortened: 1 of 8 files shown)


amymok:
@fna Push your loan_estimate.py file up as well, I am getting an import error.

test/browser_testing/features/closing_disclosure.feature
```diff
@@ -0,0 +1,61 @@
...
+@closing_disclosure
+Scenario Outline: Test that resizing window size changes image size too
+  When I click the tab "<tab_name>"
+    When I resize window image "<css_selector>" size changes too
+
+Examples:
+  | tab_name    | css_selector          |
+  | Checklist   | div.illustration img  |
+  | Definitions | img.image-map_image   |
+
```
virginiacc: I think these should both be the same selector -- `img.image-map_image`

test/browser_testing/features/closing_disclosure.feature
```diff
@@ -0,0 +1,61 @@
...
+@closing_disclosure
+Scenario Outline: Test overlays/highlights
+  When I click the tab "<tab_name>"
+    When I hover over an overlay the corresponding explainer has class has-attention
+
+Examples:
+  | tab_name    |
+  | Checklist   |
+  | Definitions |
```
virginiacc: Every time the overlay test runs, I'm getting an error for just the first item in the Checklist tab -- `#page-1-checklist-check-name`. It's failing the `has-attention` class check. If the order is affecting the results, I wonder if doing a second check for the failed items would work?

fna: I'll see what I can do. Don't merge yet

test/browser_testing/features/closing_disclosure.feature
fna: Yes, both has to be the same, but I think more specific like div.illustration img.image-map_image

virginiacc:
@fna Looks great -- everything's passing for me now, so I think this is ready to merge! :+1:

PR #525, opened 2015-07-21
Fe unit tests

cfarm:
setup fixture and unit test file for form explainers.

the fixture file is using the loan estimate HTML.

Also added .editorconfig to keep our JS formatting consistent, see https://github.com/cfpb/front-end/blob/master/javascript.md#spacing for info on how this works.


virginiacc:
:+1:

PR #529, opened 2015-07-24
Loan Comparison: auto update

virginiacc:
### Changes
- Call api automatically for new rates/insurance data when loan is updated
- Update tests to reflect this change

### Review
@fna or @stephanieosan or @cfarm or @amymok


fna:
Looks safe to merge

PR #536, opened 2015-07-28
Change innerHTML to innertext per lfatty advice

fna:
I couldn't find a way of really testing it. The page is working, data is being saved, all is functional, but it looks to me that that particular function is not being called.


virginiacc:
I don't see that this code is actually being called either. Let's make this change to fix the tests, and we can revisit if/when we decide to release the prepare worksheets. :+1:

PR #544, opened 2015-07-30
Loan Comparison: Separate interest rate and number each steps

amymok:
### Changes
- Separating out the interest rate from the rest of the inputs
- Adding numbered headers to the 4 sections (inputs, interest rate, outputs, next steps)
- Removing numbers from the input categories (about you, about the home, about the loan)

### Reviews
@stephanieosan @cfarm @virginiacc 

Questions for @stephanieosan:
* Did you want "interest rates" to be linked to somewhere under "2. choose interest rates in your scenarios"?
* In output, you see how the green bleeds to the left edge, do you really want that there?  Esp if we have big screen, it looks kind of odd

### Screenshots
screen shot 2015-07-30 at 2 14 07 pm

screen shot 2015-07-30 at 2 14 20 pm

screen shot 2015-07-30 at 2 14 29 pm

### DIFF
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -1062,7 +1062,7 @@ table.unstyled {
 
 }
 
-.app-container {
+.with-link {
     & a {
     border-width: 0;
     border-style: dotted;
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -1059,4 +1059,15 @@ table.unstyled {
   float: left;
   margin-top: -3px;
   margin-right: 10px;
+
+}
+
+.app-container {
+    & a {
+    border-width: 0;
+    border-style: dotted;
+    border-color: #0072ce;
+    color: #0072ce;
+    text-decoration: none;
+  }
 }
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -677,10 +677,6 @@
   }
 }
 
-.outputs-heading {
-  margin-left: @grid_gutter-width*2;
-}
-
   .results-section-heading {
     margin: 0;
     > .cf-icon {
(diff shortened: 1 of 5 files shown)


stephanieosan:
@amymok 
We want the copy "interest rates" to link to consumerfinance.gov/check-rates. 

We do *not* want the green highlight to extend off screen like that, but since we're removing the prepopulated mode (which includes the green highlights), that should take care of itself?

amymok:
@virginiacc @cfarm : Is this a true statement?

"We do not want the green highlight to extend off screen like that, but since we're removing the prepopulated mode (which includes the green highlights), that should take care of itself?"

amymok:
@cfarm @virginiacc When you are done with FE, I need some help getting the link to have dotted underline.  Can't seem to get this working.

src/static/css/module/loan-comparison.less
```diff
@@ -1052,3 +1048,26 @@ table.unstyled {
...
 }
 
+/* Title Heading */
+.round-step {
+  width: 1.5em;
+  height: 1.5em;
+  border-radius: 1.5em;
+  line-height: 1.5;
+  text-align: center;
+  background-color: @green-tint;
```
virginiacc: I think we could use `.icon-round` as a mixin here instead of these first six lines...

src/static/css/module/loan-comparison.less
```diff
@@ -1052,3 +1048,26 @@ table.unstyled {
...
+  float: left;
+  margin-top: -3px;
+  margin-right: 10px;
+
+}
+
+.with-link {
+    & a {
+    border-width: 0;
+    border-style: dotted;
```
virginiacc: We can add the dotted underline with `border-bottom-width: 1px;`

virginiacc:
This looks great! :+1: I had a few suggestions for the css, but they're really small & I'm fine with logging them here for future reference and pulling this in as is. Let's discuss whether we want to make any changes tomorrow, and if not I'll go ahead and merge this.

PR #545, opened 2015-07-30
Form explainer: stop using deprecated .load() method

virginiacc:
### Changes
- use .on() instead of .load()

### Review
- @cfarm


cfarm:
Apparently the .on() method is also problematic :disappointed:

PR #547, opened 2015-07-30
Loan Comparison: Output Interaction when API is loading

amymok:
### Additions
- fading out the outputs when API is being called/data is loading
- fetching data message over top of faded outputs

### Reviews:
@stephanieosan I can't really take a screenshot because it happens so fast, but I can share with you live after scrum
@virginiacc @cfarm

### DIFF
--- a/src/static/css/module/loan-comparison.less
+++ b/src/static/css/module/loan-comparison.less
@@ -199,6 +199,7 @@
     animation: fadein .5s;
   }
   
+
   .updating .interest-rate-loading {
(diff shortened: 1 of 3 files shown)


amymok:
Okay I was able to take some, @stephanieosan:

screen shot 2015-07-31 at 11 45 20 am
screen shot 2015-07-31 at 11 45 20 am

virginiacc:
Looks great!  :shipit:

PR #552, opened 2015-08-04
Fix form explainer js loading bug

virginiacc:
### Changes
- Remove form-explainer.js from browserify task so it is only used where required as a dependency

### Review
@cfarm


cfarm:
that easy after all??? sweet :tada:

PR #555, opened 2015-08-05
Form explainer fix: IE8 highlights

virginiacc:
### Changes
- Update definitions highlight filters to use new teal

### Additions
- Add `.has-attention` (and hover state?) highlight filters for IE8

### Preview
screen shot 2015-08-05 at 7 31 32 pm

screen shot 2015-08-05 at 7 49 18 pm

### Testing
This will need to be tested in Saucelabs. Open loan estimate in IE8 & see darker highlight on image map when associated explanation is hovered over/clicked.

### Notes
I'm having trouble seeing the hover state for the image map links. Not sure if this is Saucelabs or another bug.

### Review 
@cfarm

### DIFF
--- a/src/static/css/module/form-explainer.less
+++ b/src/static/css/module/form-explainer.less
@@ -60,11 +60,22 @@
 }
 
 .lt-ie9 .image-map_overlay__checklist {
-    background:none;
+   background:none;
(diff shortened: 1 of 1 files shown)


cfarm:
:+1:

PR #556, opened 2015-08-06
add missing files

fna:
Don't know whether this will be enough, but w/o these two tests don't run


OrlandoSoto:
I think this should at least get your browser tests running again :+1:

cfarm:
I got this into the `form-explainer-release` and ran smoke_testing and the fe feature files, all worked as expected locally. :+1:

PR #557, opened 2015-08-06
Fix issue with link not wrapping

Scotchester:
This link was running off the edge of the screen at all widths under about 1200, so I removed the `icon-link__no-wrap` class, which seemed to be serving no purpose.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -262,7 +262,7 @@
 
           <ul class="list__unstyled list__spaced">
             <li class="list_item">
-                <a class="icon-link__no-wrap jump-link jump-link__download" href="http://www.consumerfinance.gov/f/201503_cfpb_your-home-loan-toolkit-web.pdf" target=\"_blank\">
+                <a class="jump-link jump-link__download" href="http://www.consumerfinance.gov/f/201503_cfpb_your-home-loan-toolkit-web.pdf" target=\"_blank\">
                   <span class="jump-link_text">Your home loan toolkit: a step-by-step guide</span>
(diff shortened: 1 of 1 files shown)


cfarm:
That class is so that the icon doesn't wrap to a new line, so we might want to keep it for that reason. do you have a screenshot of the issue this fixes?

amymok:
@Scotchester did you see @cfarm comment here?

Scotchester:
Sorry, I did not. Thanks, @amymok 

Here's a couple screenshots:

screen shot 2015-08-31 at 17 23 29

screen shot 2015-08-31 at 17 23 51

That may be what the class was intended to do, but it's not doing so in this case, since nothing's wrapping at all.

amymok:
@virginiacc I saw similar problems earlier and tried to fix that by just removing the `__no-wrap` from `icon-link__no-wrap`, instead of how @Scotchester done it by removing `icon-link__no-wrap` entirely. It seems to look ok in different screen sizes.  I know when I talk to @cfarm before, she wants to make sure it doesn't affect anything else such as having the icon by itself on its own line.  I don't think I see that as an issue.  Please let me know what you think here.  Thanks!

Scotchester:
Also, in my testing, I didn't see the icon on a line by itself at any size.

virginiacc:
@amymok @Scotchester This looks good to me! I think this link is too long relative to its container to use that class unconditionally, & I'm also not seeing a case where the icon wraps to the next line by itself.

PR #559, opened 2015-08-07
Form explainer: update scroll-to position for expandables

virginiacc:
Addresses issue raised in GHE 1059. 

When a form explainer image map link is clicked, the associated expandable is opened & scrolled into view, and any other open expandable is closed. The problem seems to be that the scroll-to position is calculated from the targeted expandable's position *before* the expandable animations take place -- so if an expandable above the targeted one is collapsed in this process, we'll wind up scrolling too far down the page & part of the targeted expandable will be offscreen.

## Changes
- When form explainer image map link is clicked, subtracts content height of any previous open expandable from targeted expandable's offset top to determine scroll-to position

## Testing
1. Open first expandable on form explainer page. 
2. Click the image map link for an expandable further down the page.
3. Make sure that the targeted expandable is not partially offscreen when scrolled into view.

## Preview
This is the problematic positioning logged in issue 1059:

5f2b8bc4-3cfc-11e5-9983-e33b6eaaf6d5

This is the positioning you should see:

screen shot 2015-08-07 at 3 03 05 pm

## Notes
It's hard to test the scrolling behaviour on Saucelabs due to streaming lag. If it turns out that this is problematic, an alternative solution would be to delay scrolling to the targeted expandable if there's an open previous expandable until we estimate that the collapsing animation has completed.

## Review
@cfarm
@sonnakim

### DIFF
--- a/src/static/js/modules/form-explainer.js
+++ b/src/static/js/modules/form-explainer.js
@@ -407,12 +407,29 @@ $(document).ready(function(){
   // the page until it is in view.
   $WRAPPER.on( 'click', '.image-map_overlay', function( event ) {
     event.preventDefault();
+    var offset = 0;
     var itemID = $( this ).attr('href');
-    $.scrollTo( $(itemID), {
+    var $targetExpandable = $(itemID);
+    
+    // If there's an expandable open above the targeted expandable,
+    // it will be closed when this expandable opens, so we need
+    // to subtract its height from this expandable's offset in order
+    // to get a position to use in scrolling this expandable into view
+    // ALTERNATIVE, if this is buggy: if there's an expandable open above,
+    // just delay the scroll until we estimate the expandable has closed (maybe 500?)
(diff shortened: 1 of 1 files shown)


cfarm:
I checked issue 1059 but it didn't say so I'll ask: what browser does this affect?

stephanieosan:
I think this is on more than one browser? Sonna found it while testing on content, which probably means IE10 through citrix.

mthibos:
I get the same issue on [out of date] Chrome on Dell.

sonnakim:
Yes, for me it was IE10 on Citrix.

virginiacc:
I think you can see it on most/all browsers -- it's just more noticeable on older ones. To see it on a modern browser, you might need to reduce the browser width and/or click between areas at the top & bottom of the image map a couple of times.

cfarm:
Cool, I confirmed that this is an issue in Chrome for me and that this branch fixes that issue :+1: reviewing code now

cfarm:
Looks good and works great :+1:

PR #562, opened 2015-08-10
Form Explainer - Pagination Browser Tests

amymok:
### Additions
* Browser Tests for Form Explainers (Both Closing Disclosure and Loan Estimate):
  * Check to make sure clicking on each page link will yield the correct page being displayed
  * Click Next Page Button will yield next page
  * Click Previous Page Button will yield previous page

### Reviews
@cfarm @virginiacc @fna

### DIFF
--- a/test/browser_testing/features/pages/closing_disclosure.py
+++ b/test/browser_testing/features/pages/closing_disclosure.py
@@ -124,7 +124,6 @@ class ClosingDisclosure(Base):
         elements = self.driver.find_elements_by_class_name('form-explainer_page-link')
         for element in elements:
             if element.get_attribute('data-page') == page_num:
-                print "Clicked %s" % page_num
 
                 script = "arguments[0].scrollIntoView(true);"
                 self.driver.execute_script(script, element)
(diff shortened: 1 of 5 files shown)


test/browser_testing/features/pages/closing_disclosure.py
```diff
@@ -107,8 +114,40 @@ def hover_an_overlay(self):
...
+                                     explainer_element.get_attribute('class').split(' '))
                     if 'has-attention' not in classes:
                         bad_elements += 1
 
         return bad_elements
+
+    def click_page(self, page_num):
+        elements = self.driver.find_elements_by_class_name('form-explainer_page-link')
+        for element in elements:
+            if element.get_attribute('data-page') == page_num:
```
virginiacc: This is totally optional, but I was wondering if we could use the link's text instead of its `data-page` attribute? We might change the implementation so the links no longer need a data attribute, but their text value shouldn't change.

amymok:
@virginiacc Changed to use text instead

virginiacc:
@amymok This looks great! :+1: All the tests are passing for me when run against build!

stephanieosan:
Is the goal to get these browser tests into the release tomorrow, or are we waiting for the next time we push to production to add these in?

PR #571, opened 2015-08-13
JS build process

virginiacc:
This is a potential approach to simplify the JS build process by automatically adding shared javascript (for now, the nemo code) to the base template, and optionally adding page-specific js via a new `page_js` variable.

### Changes
- remove the `entries` entry from `factor-bundle` -- this seems to be unnecessary?
- add `base.js` file that requires the nemo code
- remove nemo requirements from all page-specific js, and remove the page-specific files that were only adding nemo
- conditionally add page-specific js in base template
- use different variable for page-specific js to avoid potential conflict with sidebar `active-page` usage

### Testing
- Check that page-specific js is still loaded on `/check-rates`, `/loan-comparison`, `/loan-options`, and `/process` (did I miss anything?)
- Check that the mobile menu still works on all OAH pages

### Review
@cfarm @amymok 

### Notes
- I'm submitting this to start discussion; it doesn't need to be merged right away.


src/_layouts/base.html
```diff
@@ -80,8 +80,15 @@
...
 {% include "nemo_subnav.html" %}
 {% include "nemo_footer.html" %}
+
+<!-- Common js -->
 <script src="{{ url_for('static', filename='js/main.js') }}?v1"></script>
+<!-- Adds nemo js -->
+<script src="{{ url_for('static', filename='js/base.js') }}?v1"></script>
+<!-- Page-specific js -->
+{% if page_js %}
 <script src="{{ url_for('static', filename='js') }}/{{active_page}}.js?v1"></script>
```
cfarm: active_page should be page_js

cfarm:
This makes a lot more sense than our current setup; I will try it out locally.

cfarm:
This is working great and I'm all for it ahead of the Journey.

cfarm:
@amymok can you test this out locally too, since it's such a major change we should double triple quadruple check it. :)

cfarm:
@virginiacc Let's get this branch updated and then merge, I just noticed this error on the sources page which will be fixed by this 

screen shot 2015-08-27 at 6 06 31 pm

amymok:
Sorry, was late on this!  It looks good, but haven't test it out locally.  Will try it out when merge conflicts are resolved.

virginiacc:
@amymok @cfarm The conflicts are resolved now, so if everything looks ok we should be able to merge this...

Gruntfile.js
```diff
@@ -112,18 +112,14 @@ module.exports = function(grunt) {
...
-          //'./src/static/js/modules/loan-comparison.js',
-          //'./src/static/js/modules/prepare-worksheets/prepare-worksheets.js',
           './src/static/js/modules/loan-estimate.js',
           './src/static/js/modules/closing-disclosure.js',
           './src/static/js/modules/process.js',
-          './src/static/js/modules/home.js',
-          './src/static/js/modules/loan-options-subpage.js',
-          './src/static/js/modules/mortgage-closing.js',
-          './src/static/js/modules/mortgage-estimate.js'
+          //'./src/static/js/modules/loan-comparison.js',
```
amymok: Is this commented out on purpose?

Gruntfile.js
```diff
@@ -112,18 +112,14 @@ module.exports = function(grunt) {
...
+          //'./src/static/js/modules/prepare-worksheets/prepare-worksheets.js'
```
amymok: Is this commented out on purpose as well?

src/loan-comparison/index.html
```diff
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% set active_page = "loan-comparison" %}
+{% set page_js = "loan-comparison" %}
```
amymok: I know you put this here, but we have commented out the loan-comparison.js.  Also if you want to add page_js, you should also add it to prepare-worksheets.js

virginiacc: @amymok Thanks for catching that! I added the page variable to the prepare worksheets page.

amymok:
Looks good to me.  I also understand those other commented out js files are for tools that are on hold, so that's okay for now.

PR #576, opened 2015-08-20
Increase test coverage

fna:


test/js/dropdown-utils_test.js
```diff
@@ -38,6 +38,40 @@ describe('Dropdown utils', function() {
     expect($('option')).to.have.length(4);
   });
 
+  it('should not add options with same values', function() {
```
amymok: If options have the same values, doesn't option length should be 1 instead?

fna: There's one option already, before I call addOption twice. It should be in before or beforeEach

test/js/dropdown-utils_test.js
```diff
@@ -38,6 +38,40 @@ describe('Dropdown utils', function() {
...
+    expect( $('#foo')[0].selectedIndex ).to.equal(1);
+  });
+
+  it('should create option with value/name of ""', function() {
+    dd('foo').addOption(null);
+    $('#foo')[0].selectedIndex = 1;
+    expect($('#foo').val()).to.equal('');
+  });
+
+  it('should disable several options', function() {
```
amymok: Maybe something like it should disable selected options

fna: I don't mind renaming, but it makes disabled the options I provide values of, it doesn't matter whether those are selected or not. Feel free changing wordings if you think mine are wrong.

test/js/dropdown-utils_test.js
```diff
@@ -38,6 +38,40 @@ describe('Dropdown utils', function() {
...
+    expect($('option:disabled').length).to.equal(2);
+  });
+
+  it('should select several select elements', function() {
+    $('body').append('<select id="foo1"><option value="baz1"></option></select>');
+    $('body').append('<select id="foo2"><option value="baz1"></option></select>');
+    dd(['foo', 'foo2']).disable();
+    expect($('select :disabled')).to.have.length(2);
+    dd(['foo', 'foo1', 'foo2']).enable();
+    expect($('select :enabled')).to.have.length(3);
```
amymok: You may also want to check that :disabled now have length 0

test/js/dropdown-utils_test.js
amymok: Ah okay, I was only reading this snippet.

test/js/dropdown-utils_test.js
amymok: I was thinking in terms of you are disabling value1 and value2, and those are your selected options to disable, not necessary highlighting types of selecting.

When typing descriptions, I try to use the rules of what scenarios I am testing in this test here, what am I manipulating and what results I am expecting from it.  So I think this description here just outlines in terms of the functionalities this module provides but not the scenario it is trying to test, so it gets a bit broad.

Nothing wrong or right, just thought maybe it will provide a bit more details when reading :)

fna: You're right, that's a known problem of mine, I am just afraid that if I start giving long descriptions it's going to be hard to understand even for me. If you have a description in mind just rename it.

test/js/loan-comparison-api_test.js
```diff
@@ -0,0 +1,86 @@
...
+describe('Loan comparison API tests', function() {
+  before(function() {
+    api = require('../../src/static/js/modules/loan-comparison/api.js');
+  });
+
+  describe('Fetch county data', function() {
+    it('should fetch county data', function() {
+      var arg = {state: 'DC'};
+      api.fetchCountyData(arg);
+      // TODO: don't know yet how to spy/stub jQuery
```
amymok: Do you plan to do this at a later PR?  Or in this PR?

You can look at promise and deferred in loan-store tests, or something like here: http://stackoverflow.com/questions/13148356/how-to-properly-unit-test-jquerys-ajax-promises-using-jasmine-and-or-sinon

fna: I'll have to do that later.

test/js/loan_store_test.js
```diff
@@ -1060,16 +1214,80 @@ describe('Loan store tests', function() {
...
+        getAll = sandbox.stub(loanStore, 'getAll');
+        loanStore.addChangeListener(loanStore.getAll);
+        loanStore.removeChangeListener(loanStore.getAll);
+        loanStore.emitChange();
+        sinon.assert.notCalled(loanStore.getAll);
+      });
+    });
+
+    describe('dispatch token', function() {
+      it('should do something magical', function() {
```
amymok: ?!

test/js/scenario-store_test.js
```diff
@@ -0,0 +1,44 @@
...
+  beforeEach(function() {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  describe('get scenario', function() {
+    it('should return the correct scenario and save the xx century fox studio', function() {
```
amymok: I don't understand this description

test/js/scenario-store_test.js
```diff
@@ -0,0 +1,44 @@
...
+      expect(result['val']).to.equal("downpayment");
+    });
+  });
+
+  describe('add change listener', function() {
+    it('should add a change listener', function() {
+      var stub = sinon.spy(scenarioStore, 'getScenario');
+      scenarioStore.addChangeListener(scenarioStore.getScenario);
+      scenarioStore.emitChange();
+      sinon.assert.calledOnce(scenarioStore.getScenario);
```
amymok: I am not sure how you can check add change listener...but this assertion doesn't really check that.

test/js/dropdown-utils_test.js
amymok: I dunno if I can push to your repo..but here's my attempt to do it, still not the best....
https://github.com/fna/owning-a-home/compare/increase-test-coverage...amymok:fna-increase-test-coverage

amymok:
Can you also use the pull request template for your PR here?

test/js/loan_store_test.js
fna: Writing tests is very hard, not knowing English doesn't help

test/js/scenario-store_test.js
fna: Was bored, I'll edit it

test/js/loan_store_test.js
amymok: I don't think it is a good idea to execute the function under test when we are not asserting anything.  It gives false data on test coverage.  So probably best to comment out the function under test so at least we know we haven't test them and will go back to them.

I think I saw that on other tests too, I am going to try to find them all.

amymok:
Looks good.

PR #591, opened 2015-08-25
wp remove index

amymok:
This is to remove passing in index so it doesn't override the index when doing sheer reindex.


amymok:
@cfarm Sorry for being bad on doing this PR in a non-standard way.  Basically the line that was removed will override the setting we have for index, so it will cause journey to look at an incorrect index.

Please add to this if I misspoke @rosskarchner

PR #593, opened 2015-08-25
Rate Checker: Fix formatting for links

amymok:
Some links on Rate Checker page look really awkward, we try to fix these to make them look better

## Changes

- Change go-link class to jump-link jump-link__right for some links

## Review

- @stephanieosan @mthibos @cfarm @virginiacc 

## Screenshots
It looks like this now if the screen size makes everything wrapped
screen shot 2015-08-25 at 1 54 50 pm

But at some point it may look like this where the lonely > is by itself if the screen is a little bit smaller
screen shot 2015-08-25 at 1 55 00 pm

Mobile looks like this:
screen shot 2015-08-25 at 1 55 24 pm

## Checklist

* [X] Changes are limited to a single goal (no scope creep)
* [X] Code can be automatically merged (no conflicts)
* [X] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Visually tested in supported browsers and devices


amymok:
@cfarm @virginiacc 

This is what I notice:

IE11, the 2nd link doesn't have a line:
screen shot 2015-08-25 at 6 40 12 pm

IE8: The number for each step doesn't work/increment:
screen shot 2015-08-25 at 7 10 04 pm

Are these known issues?

virginiacc:
@amymok Those are some weird bugs!

Not sure what's going on with the IE11 links -- we could bring it up in cf-typography.

I think the IE8 number issue is caused by [the bug described here](http://fairwaytech.com/2012/02/the-ie8-hover-bug-the-most-awesome-ie-bug-ever/) (we're using counters to generate the numbers, and the affected tab is hidden on page load). The summary, from that post:

- IF you are using IE8 (the native version, not IE9 in IE8 compatibility mode!),
- AND you are using the CSS ordered list numbering approach described above,
- AND the html that has the classes that use the “counter-reset” and “counter-increment” CSS attributes is HIDDEN when the page loads,
- THEN whenever that hidden html is displayed,
- ALL of the automatic numbers will be ZERO,
- BUT ONLY IF THE CSS :hover PSEUDO-CLASS IS USED ON THAT PAGE!

To fix this, we could look at hard-coding the numbers instead of using counters, or maybe positioning the tab content off-screen so it's not hidden on page load.

Other than that, this looks good! I think we can fix the IE numbers separately since that's an existing issue on cf.gov, but we might want to investigate the IE11 links more before merging...

virginiacc:
@amymok Looking around OAH in IE11, I see that the link border issue is showing up on most pages:

screen shot 2015-08-27 at 10 33 36 am

So I think we can merge this and then pursue both IE issues separately.

@Scotchester, have you seen this IE11 behavior? Should we log an issue about it in cf-typography?

Scotchester:
@virginiacc Huh. Wow. Haven't encountered that before, but can't honestly say I've tested a whole lot in IE11, considering IE's been pretty darn reliable since 10.

It'd _probably_ be a cf-core, issue, though, as it's (likely) related to link underlines being created with dotted borders. Which, I promise, I'm going to rectify soon...

Can we live with this for a little while? If not, I'll take a look at finding a temporary solution.

cfarm:
agree these CSS issues need to be addressed in cf-core/cf-typography

@amymok this branch has conflicts now

amymok:
@stephanieosan @huetingj 
See above on IE11 issue, can we live with that for a while?

amymok:
@Scotchester I put in the following issue for IE11: https://github.com/cfpb/cf-core/issues/43

amymok:
@virginiacc @cfarm I think this is ready for merging again!

virginiacc:
Looks good! :+1:

PR #606, opened 2015-09-01
Monthly payment worksheet pdf improvements

stephanieosan:
### Changes
* Corrected tab order (should tab though columns now instead of rows)
* Other tab behavior should work, at least in acrobat (tab on to next page, skip over the unfillable fields)
* Updated the link to the sources page to include www
* Messed around with the positioning of the blue form field overlays so hopefully the underlines look more consistent

### Notes
* This does **not** fix the problem of the prepopulated zeros not automatically deleting. I will look into that more.
* This does **not** fix any of our browser + javascript pdf troubles, since I don't think we've agreed on what we'll do as a team.

### Review 
@virginiacc 
cc: @mthibos 

Virginia, up to you whether you want to pull this in before you deploy.


stephanieosan:
Added copy to the top of the page to indicate users should download the PDF.
screen shot 2015-09-03 at 3 38 12 pm

virginiacc:
@stephanieosan The tabbing order looks great in the file from the first commit, but it seems like it might have changed back in the last commit?

stephanieosan:
Hmmmmm, weird. Let me look into it.

stephanieosan:
@virginiacc I fixed the tab order again, check it out!

virginiacc:
@stephanieosan Looks great! :+1:

PR #613, opened 2015-09-09
Monthly payment worksheet update to intro message

stephanieosan:
### Changed
* Updated new copy in alert
* Added text link
* Added Acrobat logo and link

### Screenshot
screen shot 2015-09-09 at 12 56 39 pm

### Review
@sonnakim and @virginiacc


virginiacc:
@stephanieosan  Looks good! :+1: 

The only issue I see is that the link to download Adobe Reader is causing an error in Chrome -- I think if we change the protocol from `https` to `http` it should work...

mthibos:
According to my internet, the link is: https://get.adobe.com/reader/

If I type it in without the *s*, it just redirects me to https.

virginiacc:
@mthibos Yeah, that's definitely the correct link, but I'm getting a privacy error from Chrome when I click it in the PDF. It could be something specific to my computer, though -- anyone else seeing this? @amymok

mthibos:
I can't seem to download the file, only see an image of it, so I can't check.

stephanieosan:
I'm getting a privacy pop up too. I'll update so they go to http.

stephanieosan:
Okay, should be updated. Let me know if that still is opening a privacy pop up (it isn't anymore on my computer).

virginiacc:
@stephanieosan Works perfectly! :+1:

PR #630, opened 2015-09-16
New math-less PDF

stephanieosan:
### Changes
* New monthly payment worksheet minus all the buggy math and with design improvements

### Review
* @virginiacc or @amymok 

cc: @huetingj and @mthibos


huetingj:
:+1:

PR #646, opened 2015-11-05
Adding Planning for Retirement to main nav

Scotchester:
Should not deploy to production until 11/12 at 9:00 am EST.

### DIFF
--- a/src/_layouts/nemo_subnav.html
+++ b/src/_layouts/nemo_subnav.html
@@ -21,6 +21,7 @@
       <li><h5><a href="/askcfpb/">Ask CFPB</a></h5>
         <p>Get answers to your financial questions.</p></li>
       <li><h5><a href="/paying-for-college/">Paying for College</a></h5></li>
+      <li><h5><a href="/retirement/">Planning for Retirement</a></h5></li>
       <li><h5><a href="/owning-a-home/">Owning a Home</a></h5></li>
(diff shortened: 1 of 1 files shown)


cfarm:
@virginiacc any complications with this release plan considering we have monthly-payment-worksheet in master? I would think it's ok so long as it's not been deployed via the grunt task yet.

Scotchester:
Oh, sorry I forgot to do this already, but I was going to file a new PR that doesn't put this into `master`, following our conversation yesterday, @cfarm. But if you think it's okay this way, let me know.

virginiacc:
@cfarm Hmm, good question. It's possible the worksheet will be ready to release on 11/12, but if not we should still be able to deploy from master since we're using that grunt task as a feature toggle. We'd probably want to comment out the worksheet js file in the browserify build task, though. 

We also need to specify a [version for the jquery sticky plugin](https://github.com/cfpb/owning-a-home/pull/647) next time we release to avoid breaking the form explainers, so we'd want to make that change in the release branch if we don't deploy from master.

Scotchester:
Ok, then, if this is all good, merge when ready and we'll get it on Build and Content!

Scotchester:
Friendly reminder, @cfarm and @virginiacc :)

virginiacc:
Looks good! :+1:  Thanks, @Scotchester!

PR #654, opened 2015-12-17
Added a 2 second sleep to prevent timeout errors

OrlandoSoto:
We should probably find a better way to handle these errors in the future but this is all I can do right now


OrlandoSoto:
Reviewer: @amymok , @fna

amymok:
Looks good

PR #662, opened 2016-05-18
Replace jshint with eslint

anselmbradford:
## Changes
- Replaces jshint with eslint and updates codebase to cfgov-refresh standards.
- Fixes stylistic and obvious errors, **leaves errors that need closer attention**.

## Test
- `npm install && grunt && grunt lint` Output should be `435 problems (28 errors, 407 warnings)`.

## Review
- @Scotchester
- @amymok  
- @virginiacc 
- @cfarm


anselmbradford:
Ping!

anselmbradford:
@virginiacc @cfarm Do you have time to review this? It's a pain to fix conflicts on so it would be great to get in if it looks good to you.

src/static/js/modules/explore-rates.js
```diff
@@ -563,362 +562,370 @@ function checkForJumbo() {
...
+    $( '#hb-warning' ).addClass( 'hidden' );
     // Select appropriate loan type if loan was kicked out of jumbo
     if ( params.prevLoanType === 'fha-hb' ) {
-      $('#loan-type').val( 'fha' );
-
-    }
-    else if ( params.prevLoanType === 'va-hb' ) {
+      $( '#loan-type' ).val( 'fha' );
+    } else if ( prevLoanType === 'va-hb' ) {
+      $( '#loan-type' ).val( 'va' );
```
virginiacc: I think these lines are unnecessary -- the next condition runs the same check but accesses `prevLoanType` on the params object.

src/static/js/modules/explore-rates.js
```diff
@@ -239,91 +251,85 @@ function renderLoanAmount() {
...
+    options.request.abort();
   }
 
   // And start a new one.
-  if ( +params['loan-amount'] !== 0 ) {
-    options['request'] = getData();
-    
-    
+  if ( +Number( params['loan-amount'] ) === 0 ) {
+    resultWarning();
```
virginiacc: Do we still need the plus operator if we're calling Number()?

virginiacc:
@anselmbradford This looks good to me! I'm getting a very similar output from grunt lint, and all the pages I've been able to check seem to be functioning correctly. I haven't gotten the `process` pages to run locally yet -- they're throwing a `ConnectionError`-- but the changes to `process.js` look like they only involve spacing.

src/static/js/modules/explore-rates.js
anselmbradford: Hmm, I'm not sure.. this update was following what the linter told me to do. I'd be inclined to merge it to get eslint in and then experiment on whether this would be needed or not. Nice find!

PR #667, opened 2016-06-17
Fix explore rates dropdown sort order

virginiacc:
The sort order of the interest rates in the dropdowns is sometimes wrong because the labels are being unnecessarily re-sorted once they've had percentage signs added (See OAH notes 1114 for more info)
## Additions
- A `unique` function to replace the jQuery unique method (since it also sorts)
## Changes
- Stop sorting in the updateComparisons function (labels are already sorted at this point)
## Testing
- checkout branch and run `frontendbuild.sh`
- go to [explore rates](http://localhost:7000/explore-rates/)
- Update form: 
  - Choose `California` from state dropdown
  - Set house price to `900,000`
  - Set downpayment to `20%`
  - Choose `Alameda` from county dropdown
- Check that rates in the dropdowns are correctly sorted
## Review
- @cfarm 
## Screenshots

**Before, incorrectly sorted dropdown (on `cf.gov`):**
screen shot 2016-06-17 at 8 08 00 am

**After, correctly sorted dropdown:** (the values are slightly different because the rates api returns different values on build)

screen shot 2016-06-17 at 8 12 31 am
## Checklist
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
- [ ] Passes all existing automated tests
- [ ] New functions include new tests
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged
- [ ] Visually tested in supported browsers and devices 
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/src/static/js/modules/explore-rates.js
+++ b/src/static/js/modules/explore-rates.js
@@ -434,7 +434,7 @@ function getCounties() {
 function loadCounties() {
 
   // And request 'em.
-  request = getCounties();
+  var request = getCounties();
   request.done(function( resp ) {
 
     // If they haven't yet selected a state highlight the field.
--- a/src/static/js/modules/explore-rates.js
+++ b/src/static/js/modules/explore-rates.js
@@ -334,7 +334,7 @@ function updateView() {
         return;
       }
 
-      data.uniqueLabels = $.unique( data.labels.slice(0) );
+      data.uniqueLabels = unique( data.labels.slice(0) );
 
       chart.stopLoading();
       removeAlerts();
@@ -367,6 +367,18 @@ function updateView() {
     downPaymentWarning();
   }
 };
+
+function unique (arr) {
+    var m = {}, newarr = []
+    for (var i=0; i<arr.length; i++) {
+      var v = arr[i];
+      if (!m[v]) {
+        newarr.push(v);
+        m[v]=true;
+      }
+    }
+    return newarr;
+}
 
 /**
  * Updates the sentence above the chart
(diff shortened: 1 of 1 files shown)


mthibos:
@virginiacc  Great sleuthing!! 

Question:  what does the Update Comparisons function do?  When does it fire?  Just want to make 100% sure that there is no need to resort after that function fires.

virginiacc:
@mthibos The `updateComparisons` function is called to update the two interest rate drop downs every time interest rates are fetched, & is passed a list of rates that have already been sorted and had percentage signs added to them. I think at one point it must have received unsorted rates because it tries to sort them, but the numerical comparison it calls fails since the rates are now strings instead of numbers. Most of the time the sort failure has no effect, but under certain circumstances it causes the rates to be re-ordered incorrectly.

mthibos:
@virginiacc  Thanks for the explanation.  Sounds good to me.  I checked this out on build and am satisfied that it has solved the problem.  Good to deploy to production!

mthibos:
@virginiacc   Wait, I take that back!  It's not completely fixed.  I still think we can deploy this code to production, as it solves most of the problem, I think.  But I think there is still a bit more work to be done to chase down all potential sources of this error.   I will document here:  GHE/OAH/OAH-notes/issues/1114

virginiacc:
@mthibos @cfarm thinks that the code hasn't been updated on build. I'm taking a look now...

mthibos:
Hrm.... it said it was merged!  And it seemed fixed for so many scenarios!  :confounded:

virginiacc:
@mthibos Yeah, it's strange -- it looks like the old javascript for explore rates is still up on build, even though the new code was merged...

mthibos:
@virginiacc  Hrm... ok well I won't spend any more time trying to test this fix or the other one until I hear from you that the code is fully updated on build.  :confused:

virginiacc:
@mthibos Ok, it should be updated on build now! No idea why it wasn't before.

mthibos:
@virginiacc   Reviewed this on production a couple of weeks ago, but failed to close out the issue.  It's fixed!  Huzzah!  :beers:

PR #674, opened 2016-07-07
added TUNNEL-INDENTIFIER env variable in order to use unique tunnel per build

OrlandoSoto:
we need this change to tell environment.py to get the tunnel name from the TUNNEL-INDENTIFIER env variable set by the sauce on demand plugin


OrlandoSoto:
pls review: @amymok @virginiacc @cfarm

amymok:
looks good!

PR #680, opened 2016-09-23
switch templates to using static() function where possible

rosskarchner:
This helps better take advantage of some bandwidth-saving measures we're putting in place, see: https://github.com/cfpb/cfgov-refresh/pull/2407

### DIFF
--- a/src/closing-disclosure/index.html
+++ b/src/closing-disclosure/index.html
@@ -19,7 +19,7 @@
         </div>
 
         <div class="content-l_col-1-3">
-          <img src="../static/img/ill-form-explainer-closing-disc.png" alt="" class="illustration">
+            <img src="{{ static('owning-a-home/static/img/ill-form-explainer-closing-disc.png')}}" alt="" class="illustration">
         </div>
       </div>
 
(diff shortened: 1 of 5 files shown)


KimberlyMunoz:
I have a question, should LESS files and React.JS files also be adjusted?

There's L136 of `cf-theme-overrides.less`
 `@grid_box-sizing-polyfill-path: '/static/vendor/box-sizing-polyfill';`

And a react file at `/src/static/js/modules/loan-comparison/components/scenario-section.js`
``

Also, the index file of Owning A Home uses its own static function. Should that be updated?
`Illustration of house keys`

rosskarchner:
- It seems like the preprocessor does the right thing for absolute references under /static/, but ideally react and less files would use relative paths (relative to the *built* location of the file)

That url function already does the right thing, though switching it to the static() function would be more explicit and less dependent on sheerlike "magic".

KimberlyMunoz:
:+1:

rosskarchner:
hmm, seems like something went crazy on that rebase

rosskarchner:
Let's try again...

PR #683, opened 2016-09-26
Add commas to explore rates tool

KimberlyMunoz:
This adds commas to the house price and downpayment field on refresh and
as people are typing.

## Testing

1. Visit http://localhost:8000/owning-a-home/explore-rates/
1. Type a five or six figure number for home price.
1. Change the downpayment as well.

## Review

-@virginiacc

### DIFF
--- a/src/static/js/modules/explore-rates.js
+++ b/src/static/js/modules/explore-rates.js
@@ -712,7 +712,6 @@ function renderDownPayment() {
   }
 
   checkIfZero( $price, $percent, $down );
-  addCommas( $el );
 
   if ( $price.val() !== 0 ) {
     if ( $el.attr( 'id' ) === 'down-payment' || options['dp-constant'] === 'down-payment' ) {
@@ -1198,7 +1197,6 @@ $( '.calc-loan-amt .recalc' ).on( 'keyup', function( evt ) {
 $( '.calc-loan-amt, .credit-score' ).on( 'keyup', '.recalc', function( evt ) {
   var verbotenKeys = [ 9, 37, 38, 39, 40 ];
   var element = this;
-
   // Don't recalculate on TAB or arrow keys.
   if ( verbotenKeys.indexOf( evt.which ) === -1 ||
        $( this ).hasClass( 'range' ) ) {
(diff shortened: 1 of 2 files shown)


KimberlyMunoz:
I'm going to look at this some more since the dropdown is being affected as well.

virginiacc:
@KimberlyMunoz Yes, good catch! I think you can just call `addCommas` on `$price` instead of `$el`. Or you could try the plugin, although it may be overkill since we already have functionality to enforce numeric input and format USD...

KimberlyMunoz:
I tried a couple different things. I tried comma-izing as the chart refreshes. Tracking the cursor changes was going to be a much bigger thing, although the autonumeric code was really helpful in almost getting it to work!

I now have it comma-ize on blur, so we don't have any cursor issues. Can you take a look, @virginiacc?

src/static/js/modules/explore-rates.js
```diff
@@ -1195,6 +1207,11 @@ $( '.calc-loan-amt, .credit-score' ).on( 'keyup', '.recalc', function( evt ) {
   }
 } );
 
+$( '.calc-loan-amt, .credit-score' ).on( 'blur', '.recalc', function( evt ) {
```
virginiacc: `$('.calc-loan-amt .recalc')` includes the `#county` select, so `addCommas` is being called when it blurs and changing its value.

KimberlyMunoz: Good call, I made this way more specific and changed the on blur to focus out since the changing on blur affects how people arrow back to change the values in e4a8d51

virginiacc:
@KimberlyMunoz Looks great! 👍

KimberlyMunoz:
Ready for @mthibos to view on build!

mthibos:
@KimberlyMunoz   I am not seeing any changes on build..... 
(also, the build API seems to be down, which will make it particularly hard to properly test this.  @fna may be able to jump start it for you).

mthibos:
@KimberlyMunoz  Just checking in on this.  What's the status?

KimberlyMunoz:
Amy and I paired to fix the API on build. Check it out!

PR #684, opened 2016-09-27
Post-Review fixes for breadcrumbs

KimberlyMunoz:
- Made the switch from content-wrapper__medium to content-wrapper to
  make sure it matches the new grid completely
- Switched breadcrumbs on loan option type pages
- Removed CSS that was for the old grid
- Double checked to make sure the sidefoot shaded grid is always flush with the bottom border

## Review
- @virginiacc

### DIFF
--- a/src/_layouts/loan_types_subpage_heading.html
+++ b/src/_layouts/loan_types_subpage_heading.html
@@ -1,6 +1,6 @@
-<div class="content_wrapper content_wrapper__medium">
+<div class="content_wrapper">
     <nav class="breadcrumbs" aria-label="Breadcrumbs">
-      <a href="/owning-a-home/loan-options" class="breadcrumbs_link">Understand loan options</a>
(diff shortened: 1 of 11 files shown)


KimberlyMunoz:
## Preview

image

This one was a pain in the neck because of the lack of a sidebar, but I would love a closer look on

image

Switched the breadcrumb order here :x 

image

KimberlyMunoz:
Also, thanks @mthibos for catching these fixes. I'll tag you again when this is merged.

mthibos:
:+1: 

@sonnakim @stephanieosan @huetingj 
Something to think about as we move forward with thinking about our options for incorporating new stuff on the subpages -- do we want to consider putting in a gray sidebar on the main loan options page?  Just an idea.

KimberlyMunoz:
This is live on build @mthibos

mthibos:
:+1:   Looks great!

PR #686, opened 2016-10-12
Additional Print Styles

KimberlyMunoz:
Made some adjustments on print styles to reduce the number of pages being printed.

Since Chrome prints with mobile styles and Firefox/IE prints on desktop styles, the styles need to be flexible enough to work with both.

These aren't perfect, but they're a huge improvement, especially for Firefox/IE where a lot of text was being cut off.

## Testing
- Review and Print Preview the following pages:
  - http://localhost:8000/owning-a-home/explore-rates/
  - http://localhost:8000/owning-a-home/closing-disclosure/
  - http://localhost:8000/owning-a-home/loan-options/
  - http://localhost:8000/owning-a-home/loan-options/FHA-loans
  - http://localhost:8000/owning-a-home/process

### DIFF
--- a/src/static/css/module/explore-rates.less
+++ b/src/static/css/module/explore-rates.less
@@ -10,6 +10,21 @@
     }
 }
 
+@media (max-width: 720px) {
+  .rc-home-illu {
+    display: none;
+  }
+  .rc-illu-inner {
+    padding: 0;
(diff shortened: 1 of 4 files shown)


contolini:
Amazing.

PR #688, opened 2016-10-18
Add enter to the list verbotten keys

KimberlyMunoz:
I've added the enter key as a verbotten key. Since there's two places
that track these keys, I've added it to the config to make it easier to
update.

I also found some code that wasn't doing much except messing with the
focus of cursor when editing numbers in the explore rates tool. 
The function on L1170 does the same as the deleted code.

## Review
- @virginiacc 

## Testing
Type in numbers in the fields. Hit enter and nothing should happen.

Edit the middle digit in one of the number-fields. (ex: Change 200,000 to 255,000.) The cursor should not reset to the end until you shift focus to another field.

### DIFF
--- a/src/static/js/modules/explore-rates.js
+++ b/src/static/js/modules/explore-rates.js
@@ -67,6 +67,7 @@ var params = {
   'isJumbo':        false,
   'prevLoanType':   '',
   'prevLocation':   '',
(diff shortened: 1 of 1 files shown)


mistergone:
👍 Works for me!

PR #689, opened 2016-10-20
Adding new color palette

KimberlyMunoz:
I've updated the color palette and updated all the colors in the project
accordingly. One legacy color still remains because it is used by the
version of Capital Framework we are using and can be safely removed when
we update the project. It has been clearly labeled as a legacy variable.
## Testing
- View all the pages and do a .frontendbuild.sh to ensure a proper build.


virginiacc:
👍 Looks good!

PR #693, opened 2016-11-03
Added a brand header for Owning a Home pages

KimberlyMunoz:
Created an include for a brand header to be included on all Owning a
Home pages. Right now this will be only on non-wagtail pages, but any
advice that can make this portable to wagtail is appreciated.

## Review
- @virginiacc 

## Testing steps
- Run `grunt`
- Test the header on all non-wagtail "owning a home" pages.
- Test along multiple breakpoints

## Todo
- [ ] Add to CHANGELOG

### DIFF
--- a/src/static/css/module/brand-header.less
+++ b/src/static/css/module/brand-header.less
@@ -4,20 +4,6 @@
   padding-bottom: 30px;
   border-bottom: 1px solid @gray-50;
 
-  &__icon {
-    color: @green;
-    font-size: 2.5em;
-    text-align: center;
-    .grid_column(2);
-
-    + .brand-header__headline {
-      .grid_column(10);
-      .webfont-medium();
-      font-size: 18px;
-    }
-  }
-
-
   &__1-2-container {
     padding: 15px;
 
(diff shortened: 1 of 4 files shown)


src/static/css/module/brand-header.less
```diff
@@ -0,0 +1,32 @@
...
+  background-color: @gray-10;
+  padding-top: 30px;
+  padding-bottom: 30px;
+  border-bottom: 1px solid @gray-50;
+
+  &__icon {
+    color: @green;
+    font-size: 2.5em;
+    text-align: center;
+    .grid_column(2);
```
virginiacc: I'm wondering about the spacing around the icon. On a small screen especially, it seems like it could be left-aligned:

screen shot 2016-11-08 at 10 51 35 pm

Could it be positioned absolutely?

KimberlyMunoz: I refactored a few things and I think it works with absolute positioning. I wanted to avoid setting the icon in the CSS to make it easier for other tools to configure. Check it out in 3454209

KimberlyMunoz:
This is ready for review, @mthibos.

@chosak, can you wait to deploy until Megan takes a look on build? Thanks!

PR #717, opened 2016-12-06
Update Google Tag Manager snippets

contolini:
The Analytics team requested we update our GTM snippets across all our sites. It now lives in the head but the `<noscript>` snippet remains in the body.


imuchnik:
:shipit:

PR #730, opened 2017-07-14
Update cfgov-sheer-templates dependency

Scotchester:
Had to update the shrinkwrap to get it to install the latest cfgov-sheer-templates (2.1.5). This will ensure that the header and footer styling is correct post-CF4 upgrade.


CHANGELOG.md
```diff
@@ -4,7 +4,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+- Update cfgov-sheer-templates dependency to 2.1.5 to support cfgov-refresh's
+  update to Capital Framework 4.
+
+## [2.0.2] - 2017-06-19
+### Changed
 - Modify `@grid_wrapper-width` to match Capital Framework
+- Update the About Us section to the updated text and call to action
+
+### Fixed
+- jQuery global reference
```
anselmbradford: What was done here?

anselmbradford:
Looks good, just be sure the Changelog changes are how you want them.

CHANGELOG.md
Scotchester: This includes changes from previous PRs that weren't added to the changelog. All I've done here is what is now in the Unreleased section.

Scotchester:
To clarify further: After merging this, I'll manually push an update to package.json and CHANGELOG.md to make this the v2.0.3 release.

sebworks:
@Scotchester, what version of NPM are you using?

Scotchester:
@sebworks 3.10.10

sebworks:
ok, just wondering why the shrinkwrap file got so small. It must be the new format.

Scotchester:
Hmm... that is kind of odd. It looks like maybe subdependencies are not being added? Should I have run something other than just `npm shrinkwrap` to create the file?

sebworks:
@Scotchester, what version of node are you running?

Scotchester:
@sebworks 6.11.0

sebworks:
ooooh ok, it's only shrinkwrapping the dependencies, which is correct. The OAH package.json file is a little different than cfgov-refresh.

In cfgov-refresh we list all our build dependencies as `dependencies` and not `devDependencies`. So `npm shrinkwrap --dev` should work.  

I'm interested to see if this will fail on Jenkins as it currently exists.

Scotchester:
Shouldn't it still include subdepdendencies of those regular dependencies, though?

sebworks:
looks good, 👍

Scotchester:
> Shouldn't it still include subdepdendencies of those regular dependencies, though?

Responding to my own query here, for the record: It actually _was_ including subdependencies, there were just very very few of them for the project's prod dependencies.

